PR #10, opened 2020-05-10
Added Firefox Android instructions for primary search settings

DjLogozzo:


benbusby:
Nice, a lot more straightforward than the process on iOS - plus it looks like it retains the POST request search functionality as well. Thanks for adding this!

PR #16, opened 2020-05-10
Add autofocus to input field

Kombustor:
Supported in all major browsers, allows the user to immediately start typing after loading the page.

### DIFF
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -28,7 +28,7 @@
             <img class="logo" src="/static/img/logo.png">
             <form action="/search" method="post">
                 <div class="search-fields">
-                    <input type="text" name="q" id="search-bar" autofocus>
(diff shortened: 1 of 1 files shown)


benbusby:
Just the one small comment. Let me know your thoughts and I'll merge accordingly.

Kombustor:
This seems to be a firefox bug. I currently don't have it installed, but [it should be supported](https://caniuse.com/#feat=autofocus).

We could focus the element by using JS (``document.getElementById('search-bar').focus()` on page load), but this is certainly not the best solution. And it didn't work for me in Chrome using the console.

Kombustor:
Does it work for you in firefox like this? @benbusby

benbusby:
Hmmm no, the updated autofocus field didn't fix it. I'll still merge this in since it seems to be working everywhere else, but I'll make a separate issue for looking into it further on Firefox. 

Thanks for contributing!

PR #22, opened 2020-05-11
Update README.md

ygsk10:


benbusby:
Good idea, didn't realize I hadn't really specified which config options can be included in the post request. Thanks for contributing!

PR #32, opened 2020-05-12
Swap out Flask's default web server for Waitress

RealOrangeOne:
Flask's dev server isn't designed for production use, in fact it's actively discouraged.

Waitress however is.

Gunicorn was also an option, but as there's multiple entrypoints for this application, and Gunicorn doesn't have a stable Python API, it wasn't viable. It's also a bit overkill for an application of this scale.

Also migrated the default port from 8888 to 5000. 5000 was used in more places. Rather than having 2 ports defined, it now just uses 1 everywhere.


benbusby:
I was wondering when someone was going to point out the flask dev server. I had in mind that it needed to change, but didn't want to use Gunicorn and didn't actually know about Waitress (which looks like a great alternative). At first glance everything looks good, I'll look at it a bit more later and try it out.

PR #33, opened 2020-05-12
Remove chmod of script in container

RealOrangeOne:
No need to chmod `whoogle-search` script inside container, It's already executable.

Tiny nitpick cleanup I noticed.


benbusby:
That's what I figured as well, but got a bug report early on from a user who claimed that they were getting a permission denied error until I added that in. I'm fine taking it back out.

PR #34, opened 2020-05-12
Use slim version of docker container

RealOrangeOne:
This massively reduces the size of the final container (330mb -> 60mb).

They're both still Linux based, and I believe both still based on Debian, but 1 is significantly smaller. Alpine-based was also an option, but isn't ideal for native modules.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.8
+FROM python:3.8-slim
 
 WORKDIR /usr/src/app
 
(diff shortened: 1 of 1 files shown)


benbusby:
That sounds like a nice improvement. I'll need to take a closer look later though, on initial build I'm now encountering issues with installing pycurl (seems like the slim package is missing libcurl4-openssl-dev and libssl-dev at least).

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.8
+FROM python:3.8-slim
 
 WORKDIR /usr/src/app
```
benbusby, line 4:
```suggestion
RUN apt-get update && apt-get install -y build-essential libcurl4-openssl-dev libssl-dev
```

Needs these packages, otherwise works great

Issue #37, opened 2020-05-12
[BUG] docker container won't do anything, no logs after branch 27 merge


edrock200:
**Describe the bug**
Just installed it and hour ago, was working fine except results in wrong language. I saw pull request 27, then saw it was merged, so I deleted the container and git clone folder and started over. The new version built but when it starts, nothing happens. Docker logs says "No logs." 
**To Reproduce**
Follow docker steps in readme

**Expected behavior**
Whoogle server listening on 8888 and allowing search 

**Desktop (please complete the following information):**
 - OS: docker on Ubuntu 18.04
 - Browser chrome android

edrock200:
I should have noted I kept the version before the merge, built that one in docker again and it runs fine.

benbusby:
I think what you're seeing is a side effect of pull request #32, which changed the default docker port to 5000 unless otherwise specified. The README was updated with this change, and I was planning on announcing the port change once I push the new image to DockerHub, but I realize that existing builds pulling from master will need to change ports. 

Could you try accessing the instance through port 5000 and see if that fixes it?

edrock200:
Thank you. I did notice that and tried it but doesn't work. Normally, when creating/starting the container the log at least shows "listening on 0.0.0.0:5000 but the latest build litterally says "No logs." It's odd. I'm about to try and build it on a second docker server to see if it's something with my setup.

edrock200:
Yep same exact issue on second box. No logs and no response on port 5000. confirmed its published to host port 5000.

Atticus-:
We don't see logs because python buffers its output. If we add a -u to line 24 of `whoogle-search` we can see the problem in the logs:

```
whoogle-search    | Serving on http://localhost:5000
```

During #32 the `--host 0.0.0.0` argument was left out, so the webserver only listens on localhost (despite docker exposing and publishing the port). The fix is to change line 24 in `whoogle-search` to:

```
python3 -um app --host "0.0.0.0" --port $PORT
```

Happy to send a PR if you'd like.

benbusby:
Was just about to comment this, but looks like you beat me to it. A PR would be great, thanks for looking into this as well.

benbusby:
Marking this as resolved with the merge of #38.

edrock200:
That worked! Thank you! Not sure if it matters but the logs throw this error now, however searches seem to work fine:

`Serving on http://0.0.0.0:5000


Serving on http://0.0.0.0:5000


[...]

PR #38, opened 2020-05-13
Fixed missing host parameter bug.

Atticus-:
Fixes issue #37. Copied from comment there:

> We don't see logs because python buffers its output. If we add a -u to line 24 of `whoogle-search` we can see the problem in the logs:
> 
> ```
> whoogle-search    | Serving on http://localhost:5000
> ```
> 
> During #32 the `--host 0.0.0.0` argument was left out, so the webserver only listens on localhost (despite docker exposing and publishing the port). The fix is to change line 24 in `whoogle-search` to:
> 
> ```
> python3 -um app --host "0.0.0.0" --port $PORT
> ```
> 
> Happy to send a PR if you'd like.

### DIFF
--- a/whoogle-search
+++ b/whoogle-search
@@ -21,5 +21,5 @@ mkdir -p $STATIC_FOLDER
 if [[ $SUBDIR == "test" ]]; then
     pytest -sv
 else
-    python3 -m app --port $PORT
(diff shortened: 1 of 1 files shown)


floptical:
Ah, I was just about to push this!!

PR #45, opened 2020-05-14
Added instructions for running with systemd

thomcatdotrocks:


README.md
```diff
@@ -93,6 +93,33 @@ pip install -r requirements.txt
 ./whoogle-search
 ```
 
+#### systemd Configuration
+After building the virtual environment, add the following to `/lib/systemd/system/whoogle.service`:
```
benbusby, line 97:
```suggestion
After building the virtual environment, you can add the following to `/lib/systemd/system/whoogle.service` to set up a Whoogle Search systemd service:
```

Super minor nitpick here, but just for anyone who is unfamiliar with systemd, I don't want them confused into thinking that this is a required step after the manual setup steps above. Probably won't happen, but doesn't hurt to just make the wording a bit more clear.

goldstar611:
I just heard about whoogle and thought I would ask on this PR where the change was introduced but why is `User=root` when the port is 5000? It seems that root privileges wouldn't be needed. Thanks!

PR #46, opened 2020-05-15
Added "I'm feeling lucky" function

icelander:
Now when a search query starts with `! ` it will redirect the user to the first search result

Signed-off-by: Paul Rothrock <paul@movetoiceland.com>

### DIFF
--- a/app/filter.py
+++ b/app/filter.py
@@ -13,6 +13,40 @@ BLANK_B64 = '''
 data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAoAAAAKCAQAAAAnOwc2AAAAD0lEQVR42mNkwAIYh7IgAAVVAAuInjI5AAAAAElFTkSuQmCC
 '''
 
+def get_first_link(soup):
+    # Replace hrefs with only the intended destination (no "utm" type tags)
+    for a in soup.find_all('a', href=True):
+        href = a['href'].replace('https://www.google.com', '')
+        
+        result_link = urlparse.urlparse(href)
+        query_link = parse_qs(result_link.query)['q'][0] if '?q=' in href else ''
+
+        # Return the first search result URL
+        if 'url?q=' in href:
+            return filter_link_args(href)
+
+def filter_link_args(query_link):
+    parsed_link = urlparse.urlparse(query_link)
+    link_args = parse_qs(parsed_link.query)
+    safe_args = {}
+
+    if len(link_args) == 0 and len(parsed_link) > 0:
+        return query_link
+
+    for arg in link_args.keys():
+        if arg in SKIP_ARGS:
+            continue
+
+        safe_args[arg] = link_args[arg]
+
+    # Remove original link query and replace with filtered args
+    query_link = query_link.replace(parsed_link.query, '')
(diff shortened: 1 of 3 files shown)


benbusby:
Awesome! Thanks so much for tackling this, I really appreciate it. I'll review as soon as I get the chance!

icelander:
Happy to help! I imagine it could be improved in future versions by adding a config option to disable it, but it's already improving my productivity a lot

app/routes.py
```diff
@@ -74,6 +74,11 @@ def search():
         except InvalidToken:
             pass
 
+    feeling_lucky = q.startswith("! ")
```
benbusby, line 77:
```suggestion
    feeling_lucky = q.startswith('! ')
```

Needs single quotes to match the rest of the python formatting

app/filter.py
```diff
@@ -58,6 +58,40 @@ def clean(self, soup):
 
         return soup
 
+    def get_first_url(self, soup):
```
benbusby, line 61: Since this reuses some of the link parsing functionality from the `update_links` method, I think it'd be better to structure this as follows:

- Change `get_first_url(self, soup)` to be a static method outside of the Filter class, i.e.:
```python
def get_first_url(soup):
    # Replace hrefs with only the intended destination (no "utm" type tags)
    for a in soup.find_all('a', href=True):
        href = a['href'].replace('https://www.google.com', '')

        result_link = urlparse.urlparse(href)
        query_link = parse_qs(result_link.query)['q'][0] if '?q=' in href else ''

        # Return the first search result URL
        if 'url?q=' in href:
            return filter_link_args(query_link)


class Filter:
...
```

- Add in the `filter_link_args` method seen in the previous suggested change:
```python
def filter_link_args(query_link):
    parsed_link = urlparse.urlparse(query_link)
    link_args = parse_qs(parsed_link.query)
    safe_args = {}

    if len(link_args) == 0 and len(parsed_link) > 0:
        return query_link

    for arg in link_args.keys():
        if arg in SKIP_ARGS:
            continue

        safe_args[arg] = link_args[arg]

    # Remove original link query and replace with filtered args
    query_link = query_link.replace(parsed_link.query, '')
    if len(safe_args) > 0:
        query_link = query_link + urlparse.urlencode(safe_args, doseq=True)
    else:
        query_link = query_link.replace('?', '')

    return query_link
```

- Then we can also update the `update_links` Filter class method to make use of this new helper function, i.e.:
```python
def update_links(self, soup):
...
            elif 'url?q=' in href:
                # Strip unneeded arguments
                query_link = filter_link_args(query_link)
                a['href'] = query_link

                # Add no-js option
                if self.nojs:
                    gen_nojs(soup, query_link, a)
...
```

___
With these changes, `routes.py` would also need another small update to import `get_first_url` separately from `filter`, and then would just be called as `get_first_url` instead of `content_filter.get_first_url`.

Let me know if you have any thoughts regarding these changes or need anything clarified. Not trying to nitpick or anything, just trying to make the most efficient use of the code and avoid any duplication.

benbusby:
Awesome, looks great! Thanks again for implementing this, it's much appreciated.

Issue #17, opened 2020-05-10
[BUG] Save config http only


gripped:
**Describe the bug**
I've installed this both in an AWS  free instance manually and also using your heroku quick deploy
At first, in the case of AWS, saving the config settings worked because I'd opened port 80 to get ssl certs with certbot. Once I closed port 80 again the config changes won't save . No biggie I can open port 80 ! My nginx settings redirect back to https anyway.
However with heroku I start with a https page eg https://mywhoogle.herokuapp.com. Saving the config settings take me to the http page http://mywhoogle.herokuapp.com where it will stay for any subsequent searches

**To Reproduce**
Steps to reproduce the behavior: AWS
1. Only allow traffic on port 443 https
2. Go to https://mywhoogle.domain.com
3. Edit settings and click save
4. Page hangs

Steps to reproduce the behavior: Heroku
1. https://mywhoogle.herokuapp.com
2. Edit settings and click save
3. Page is now http://mywhoogle.herokuapp.com

**Expected behavior**
AWS: allow save via https
heroku: remain on https after save if that's what we we on before saving.

Nice work :)

benbusby:
I pushed a new update that should address this issue. The config section on the main page now has an option to set the instance's root url. When you get a chance, try pulling in the new changes and make sure the app url is using https before saving your config.

gripped:
This does work to allow saving via https. If the Root URL: in the settings is changed to https://
However in the first instance, when first visiting via https:// the Root URL: is pre-populated with the http:// form of the url. I tested several times by restarting the heroku dyno.
Therefore someone not knowing they had to edit this to https:// does still get the hanging page on the AWS install (with port 80 blocked), when they click save. But on the heroku install the save will complete, and the session will be http until manually reverting to https.
So fixed for me, able to save via https, but not closing as I don't think it's fixed entirely ? A user shouldn't be getting dropped down to http because they didn't know to add an 's' to a url before saving. (not that I'll have any users, private search)

Still great work :)

gripped:
Sorry more http woes. Adding whoogle as a search engine in firefox and either using it as the default engine, or from the search choices displayed in the address bar dropdown results in an http search.

In ~/.mozilla/firefox/xxxxxxxx.default/search.json.mozlz4

```
{
	"_name": "Whoogle",
	"_shortName": "whoogle",
	"_loadPath": "[http]whoogle.informationhouse.co.uk/whoogle.xml",
[...]

PR #48, opened 2020-05-15
Feature: https only -- adds option to enforce https on running instances

benbusby:
Enforces https by default on Docker instances, adds an option to enforce https if running through pip/pipx.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,11 +10,14 @@ RUN mkdir $config_dir
 VOLUME $config_dir
 ENV CONFIG_VOLUME=$config_dir
 
-ARG use_https=1
+ARG use_https=''
 ENV HTTPS_ONLY=$use_https
+
+ARG whoogle_port=5000
+ENV EXPOSE_PORT=${EXPOSE_PORT:-$whoogle_port}
 
 COPY . .
 
-EXPOSE 5000
+EXPOSE $EXPOSE_PORT
 
 CMD ["./whoogle-search"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ ARG use_https=''
 ENV HTTPS_ONLY=$use_https
 
 ARG whoogle_port=5000
-ENV EXPOSE_PORT=${EXPOSE_PORT:-$whoogle_port}
+ENV EXPOSE_PORT=$whoogle_port
 
 COPY . .
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,9 @@ RUN mkdir $config_dir
 VOLUME $config_dir
 ENV CONFIG_VOLUME=$config_dir
 
+ARG use_https=1
+ENV HTTPS_ONLY=$use_https
+
 COPY . .
 
 EXPOSE 5000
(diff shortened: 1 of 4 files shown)


benbusby:
@FoxxMD @chpatton013 (or anyone else who comes across this and is a regular Docker user), do you have any input on this new feature? It basically just forces the Flask server to reroute to HTTPS by default if running via Docker, but instances run through other methods require a flag to enforce this behavior.

I'm a bit torn since I don't really like having different deployment options behave differently by default, but it solves a problem brought up in #17 where the Flask server (believing that it's running as HTTP only) would reroute to the non-HTTPS variety of whatever URL is being used. If the site wasn't configured properly, it either would get stuck on the HTTP connection from this point on, or just hang if port 80 wasn't opened. 

Overall I think HTTPS enforcement is a good addition to solve issues like this, but I might be overlooking some cases where this would be a breaking change. Let me know your thoughts (if you have any, feel free to ignore if you're indifferent about it).

FoxxMD:
First off, it's configurable so that's good 👍

WRT deployment options I agree with you that they should be consistent. If the default using pip is not to enforce then I don't think you should enforce in docker either. IMO this is a configuration setting that should be left up to the user to use correctly. If you feel like users need guidance on using it correctly you could always add some info to the readme, however.

As far as breaking changes this (enforcing https by default) would definitely break deployments for anyone who is already running the docker and using it behind a reverse proxy (like me).

FoxxMD:
Issue adjacent -- allowing docker users to specify the port whoogle runs on via ENV variable would be much appreciated. An [unraid user](https://forums.unraid.net/topic/92360-support-foxxmd-whoogle-search/?do=findComment&comment=855506) asked me about running on port 443 in a configuration (in unraid) that doesn't allow us to map ports between container/host so they have no real recourse ATM. Would be an easy fix I think ;)

chpatton013:
Agreed with @FoxxMD; Keep it consistent and configurable. I also run all my docker apps behind a reverse-proxy, so I prefer them to do dumb HTTP by default.

Adnn @FoxxMD WRT your request for a configurable port variable, I was actually just about to open a PR to do that.

benbusby:
Great points, I hadn't thought of the breaking changes to reverse proxies yet, so I'm glad I got a second (and now third) set of eyes on this. 

I just updated this to default to HTTP, and snuck in an update to set a configurable port variable as well since they're related.

Dockerfile
```diff
@@ -10,8 +10,14 @@ RUN mkdir $config_dir
 VOLUME $config_dir
 ENV CONFIG_VOLUME=$config_dir
 
+ARG use_https=''
+ENV HTTPS_ONLY=$use_https
+
+ARG whoogle_port=5000
+ENV EXPOSE_PORT=${EXPOSE_PORT:-$whoogle_port}
```
chpatton013, line 17: AFAIK the use of `EXPOSE_PORT` on the RHS of this assignment isn't helping you. You effectively have an empty env at image build time, and you haven't set any ARGs or ENVs for `EXPOSE_PORT` before this line. I think it's equivalent just to do `ENV EXPOSE_PORT=$whoogle_port`

benbusby: Ah crap, yeah I was messing around with a few things and didn't mean to keep that change. Thanks for pointing that out, will remove.

chpatton013:
:shipit:

PR #51, opened 2020-05-15
Feature: basic auth

### LINKED ISSUE #43: Login page for Whoogle
Login to Whoogle.

**Describe which parts of the project this would modify (front end/back end/configuration/etc)**
Probably both?

**Additional context**
Would it be possible to add a login page to Whoogle?
This way, when my Whoogle server is exposed to the internet it won't be hammered.
You will need to first login before you can use Whoogle.

benbusby:
Username/password can be set either as Dockerfile build arguments or
passed into the run script as "--userpass <username:password>"

Closes #43

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,11 @@ RUN mkdir $config_dir
 VOLUME $config_dir
 ENV CONFIG_VOLUME=$config_dir
 
+ARG username=''
+ENV WHOOGLE_USER=$username
+ARG password=''
+ENV WHOOGLE_PASS=$password
+
 COPY . .
 
 EXPOSE 5000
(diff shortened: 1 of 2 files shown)


ilovepancakes95:
First, I am totally fine with the basic auth implementation as opposed to a more complicated login page/user system, but is there a way to make it so it remembers/caches my login between browser launches? If I set whoogle as my default search engine in firefox for example, everytime I launch the browser I have to re-login to perform search which gets quite cumbersome after a whole day or working and searching. At the very least, I tried to get firefox to remember and autofill the username/password but it doesn't seem to work with the type of auth form whoogle uses. I want to run whoogle on a VPS but also want to protect it from unauthorized use and getting hammered by bots/other people if they find it, yet want my devices I add whoogle to easily access and keep using it seamlessly.

Second, how so we buy you a pack of beer or some pizzas? Seriously, every decent feature people keep requesting here seems to be implemented within a week, and it's much appreciated. Job well done on this project.

PR #52, opened 2020-05-15
Add Alfred Instructions

icelander:
This is why I wanted the "I'm feeling lucky" feature


benbusby:
Awesome, looks great! Thanks for adding this. I've never actually used Alfred before, but have heard a lot about it. I'll have to check it out sometime soon, seems super useful.

PR #53, opened 2020-05-15
Feature: Allow bind address override

chpatton013:
Allow the bind address to be controlled with an environment variable in the same way that the bind port can be controlled. This allows users to be more specific with which interface the application will listen on, enabling restriction to a private network connection on machines with multiple interfaces.

Also, performed a quick bash style pass (which I'm happy to revert if you don't like it):
* `set -euo pipefail` to catch programming errors early
* `builtin cd` instead of `cd` to correctly invoke a commonly-aliased command
* Double quotes around all variables to prevent accidental whitespace-expansion
* Use variable-defaults instead of an "if unset" block

### DIFF
--- a/whoogle-search
+++ b/whoogle-search
@@ -3,23 +3,22 @@
 # ./whoogle-search # Runs the full web app
 # ./whoogle-search test # Runs the testing suite
 
-SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd -P)"
+set -euo pipefail
 
-# Set default port if unavailable
-if [[ -z "${PORT}" ]]; then
-    PORT="${EXPOSE_PORT:-5000}"
-fi
+SCRIPT_DIR="$(builtin cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
 
 # Set directory to serve static content from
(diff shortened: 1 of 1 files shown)


chpatton013:
@benbusby I realize that this conflicts with #48. I'm curious what you think would be the best way to integrate the two. I can:
1. change the use of `$PORT` in the `whoogle-search` executable to `${PORT:-"${EXPOSE_PORT:-5000}"}`, or
1. change the `EXPOSE_PORT` `ENV` directive in the `Dockerfile` to `ENV PORT=$whoogle_port`

benbusby:
Looks great! Regarding your proposed solution on integrating the two PRs, I think option 1 is the better way to go. Option 2 would seemingly overwrite the enforced runtime principles for at least the Heroku build (namely the PORT variable they set on their end).

benbusby:
Conflicts can all be resolved in favor of your changes. Once the executable is updated with the new ternary port assignment I'll merge.

chpatton013:
Ah, I didn't realize that the `PORT` environment variable was used by Heroku. In that case, I agree we don't want to set it in the image.

PR #86, opened 2020-06-02
Project refactor (#85) -> master

benbusby:
* Major refactor of requests and session management

- Switches from pycurl to requests library
  - Allows for less janky decoding, especially with non-latin character
  sets
- Adds session level management of user configs
  - Allows for each session to set its own config (people are probably
  going to complain about this, though not sure if it'll be the same
  number of people who are upset that their friends/family have to share
  their config)
- Updates key gen/regen to more aggressively swap out keys after each
request

* Added ability to save/load configs by name

- New PUT method for config allows changing config with specified name
- New methods in js controller to handle loading/saving of configs

* Result formatting and removal of unused elements

- Fixed question section formatting from results page (added appropriate
padding and made questions styled as italic)
- Removed user agent display from main config settings

* Minor change to button label

* Fixed issue with "de-pickling" of flask session

Having a gitignore-everything ("*") file within a flask session folder seems to cause a
weird bug where the state of the app becomes unusable from continuously
trying to prune files listed in the gitignore (and it can't prune '*').

* Switched to pickling saved configs

* Updated ad/sponsored content filter and conf naming

Configs are now named with a .conf extension to allow for easier manual
cleanup/modification of named config files

Sponsored content now removed by basic string matching of span content

* Version bump to 0.2.0

* Fixed request.send return style

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@ __pycache__/
 *.pem
 config.json
 test/static
+flask_session/
+app/static/config
 
 # pip stuff
 build/
--- a/.gitignore
+++ b/.gitignore
@@ -3,10 +3,12 @@ venv/
 __pycache__/
 *.pyc
 *.pem
+*.conf
 config.json
 test/static
 flask_session/
 app/static/config
+app/static/custom_config
 
 # pip stuff
 build/
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,12 +1,24 @@
+from app.utils.misc import generate_user_keys
 from cryptography.fernet import Fernet
 from flask import Flask
+from flask_session import Session
 import os
 
 app = Flask(__name__, static_folder=os.path.dirname(os.path.abspath(__file__)) + '/static')
-app.secret_key = Fernet.generate_key()
-app.config['VERSION_NUMBER'] = '0.1.4'
+app.user_elements = {}
+app.config['SECRET_KEY'] = os.urandom(16)
+app.config['SESSION_TYPE'] = 'filesystem'
+app.config['VERSION_NUMBER'] = '0.2.0'
 app.config['APP_ROOT'] = os.getenv('APP_ROOT', os.path.dirname(os.path.abspath(__file__)))
 app.config['STATIC_FOLDER'] = os.getenv('STATIC_FOLDER', os.path.join(app.config['APP_ROOT'], 'static'))
-app.config['CONFIG_PATH'] = os.getenv('CONFIG_VOLUME', app.config['STATIC_FOLDER']) + '/config.json'
+app.config['CONFIG_PATH'] = os.getenv('CONFIG_VOLUME', app.config['STATIC_FOLDER'] + '/config')
+app.config['SESSION_FILE_DIR'] = app.config['CONFIG_PATH']
+app.config['SESSION_COOKIE_SECURE'] = True
+
+if not os.path.exists(app.config['CONFIG_PATH']):
+    os.makedirs(app.config['CONFIG_PATH'])
+
+sess = Session()
+sess.init_app(app)
 
 from app import routes
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,28 +1,28 @@
 from app.utils.misc import generate_user_keys
-from cryptography.fernet import Fernet
 from flask import Flask
 from flask_session import Session
 import os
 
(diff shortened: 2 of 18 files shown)


FoxxMD:
I'm on `beta` tag on dockerhub. Only config didn't work but that was to be expected. New config save/loading works well! And all the other fancy new features (autocomplete, spelling correct) are awesome! Great work!

Only issue I've seen so far is that every request to whoogle without using a named config produces a new session file. Like, every single one (autocomplete, page browsing, etc.) This is just a few minutes browsing around -- my folder is already up to 30kb of files.
image

FoxxMD:
Images also aren't loading and cause the page to hang for a bit (eventually recovers). Maybe related to this trace:

```
Traceback (most recent call last):
File "/usr/local/lib/python3.8/site-packages/cryptography/fernet.py", line 104, in _verify_signature
h.verify(data[-32:])
File "/usr/local/lib/python3.8/site-packages/cryptography/hazmat/primitives/hmac.py", line 66, in verify
ctx.verify(signature)
File "/usr/local/lib/python3.8/site-packages/cryptography/hazmat/backends/openssl/hmac.py", line 74, in verify
raise InvalidSignature("Signature did not match digest.")
cryptography.exceptions.InvalidSignature: Signature did not match digest.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 2446, in wsgi_app
response = self.full_dispatch_request()
File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1951, in full_dispatch_request
rv = self.handle_user_exception(e)
File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1820, in handle_user_exception
reraise(exc_type, exc_value, tb)
File "/usr/local/lib/python3.8/site-packages/flask/_compat.py", line 39, in reraise
raise value
File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1949, in full_dispatch_request
rv = self.dispatch_request()
File "/usr/local/lib/python3.8/site-packages/flask/app.py", line 1935, in dispatch_request
return self.view_functions[rule.endpoint](**req.view_args)
File "/usr/src/app/app/routes.py", line 32, in decorated
return f(*args, **kwargs)
File "/usr/src/app/app/routes.py", line 200, in element
src_url = cipher_suite.decrypt(request.args.get('url').encode()).decode()
File "/usr/local/lib/python3.8/site-packages/cryptography/fernet.py", line 75, in decrypt
return self._decrypt_data(data, timestamp, ttl)
File "/usr/local/lib/python3.8/site-packages/cryptography/fernet.py", line 117, in _decrypt_data
self._verify_signature(data)
File "/usr/local/lib/python3.8/site-packages/cryptography/fernet.py", line 106, in _verify_signature
raise InvalidToken
cryptography.fernet.InvalidToken
ERROR:app:Exception on /element [GET]
```

benbusby:
@FoxxMD thanks a million for testing this out already! 

Regarding the session files being generated for every request, I've actually encountered this before and it seemed to be related to my browser blocking cookies for my instance of whoogle-search (and so a new "session" was created on every api call). And since the element and text keys are tied to user sessions now, that would explain why some images aren't decrypting properly and you're seeing the encrypted search string when navigating between tabs. Any chance this might be the case with your setup?

I should probably revisit the logic behind all of this and ensure it works similarly for any browsers that don't allow setting cookies for the instance (or otherwise just aren't passing the session value between api calls properly).

FoxxMD:
Yep you nailed it. One of my extensions (Privacy Bader, ublock origin) was not letting the cookie be set but they also weren't showing anything being blocked.

benbusby:
Hmmm, that's interesting. I have both extensions active and haven't experienced cookies for Whoogle being blocked yet. Either way I plan to look into this quite a bit more before merging though.

FoxxMD:
If it helps at all I'm using this from my own domain (subdomain.mydomain.com) rather than accessing through IP

benbusby:
@FoxxMD I couldn't accurately track down a specific reason for why occasionally browsers might reject cookies from whoogle, but I updated this anyways to include fallback support for browsers that aren't allowing cookies to be set for one reason or another. Now anyone who uses whoogle but isn't able to (or doesn't want to) allow cookies for the site will trigger the "legacy" version of the app, which just uses the old method of using a static config file and a one-time generated key set. Hopefully this mitigates the issue you were seeing originally. I'll include a short write-up about these changes in the release documentation once I merge.

PR #89, opened 2020-06-17
Adding support to choose separately the langage of search and interface.

jacr13:
There are small changes to allow the user to choose separately the interface and search language.
In addition, it allows to not specify any so google choose the best. This is important for people using several languages, if you specify a specific language and make a search in another one the results are not very accurate.

Maybe a better name for the variable should be choose... I was not very inspired.


benbusby:
Nice, thanks for your contribution. Looks good on first glance, I should have some time this weekend to do a full review and merge.

Issue #35, opened 2020-05-12
"I'm Feeling Lucky" with shortcut


icelander:
**Describe the feature you'd like to see added**
If you put an exclamation point at the beginning of your search, Whoogle would get the first result and redirect you to that page securely and privately.

**Describe which parts of the project this would modify (front end/back end/configuration/etc)**
When receiving the search query, check the beginning of the string for `! `. If it's there, remove it from the search query and search Google. When processing the results, do a redirect to the first one instead of showing the search results.

**Additional context**
DuckDuckGo has the concept of [Bangs](https://duckduckgo.com/bang) which are shortcuts to specific search results. The most useful is the I'm Feeling Lucky, which is just `! `. In the future Whoogle could have other ones, or the ability to configure your own.

benbusby:
This is a really great idea, thanks for the suggestion! It doesn't sound too hard to add it in either, so I'll try to work this into the next release.

icelander:
[Submitted a PR](https://github.com/benbusby/whoogle-search/pull/46) :)

benbusby:
Merged in #46

PR #96, opened 2020-06-25
Added ddg-style !bang-operators

marvinborner:
I've just found this project and immediately fell in love with it!
I'm a long-time duckduckgo user and would love to use the famous 13.000+ !bang operators (if you don't know them, look them up - they're awesome!). While there is the "I'm feeling lucky"-bang already (#35), all the other operators are still missing.
I've created this pull request as a kind of proof of concept and it's definitely not mergeable yet. It uses hardcoded operators and the functions are scattered around in the code (I'm not that comfortable with the codebase, yet). The best-case-scenario for me (personally) would be the ability to use all of the ddg !bang operators, and I'm sure that many others would love this, too.

Basically, this is a general discussion whether this feature should get implemented and how exactly this could be accomplished without having a giant list of operators in the source-code.

Thanks for considering the implementation of this feature.


benbusby:
Hi @marvinborner, thanks for making this suggestion -- I think it's a great idea. As far as a good way to approach this issue, my initial reaction is that it seems like it'd be a better idea to just directly consume the DDG API conditionally if the query begins with `!`. That way all of the available bang operations are directly available to Whoogle users without requiring any large modifications to the existing backend. 

[Their API seems to be designed for this use case anyways](https://duckduckgo.com/api), with a few important requirements (namely that they're attributed for this feature, which I'm thinking could be implemented in the UI as part of the autocomplete dropdown whenever a bang operator is being used). 

I haven't tried this out yet, but seems like it should be fairly straightforward to implement -- though it'd be nice if this used a new API route instead of `/search` just to keep things as separate/easy to test as possible. Let me know your thoughts!

marvinborner:
I thought of using the DDG API too, but it seems suboptimal regarding speed and privacy. While I don't know if it's allowed, one possibility would be to crawl the DDG API for bang operators once and create an extensive key-value based list (using something easily parseable like csv).
No matter whether the API or the list is used, your other ideas sound great and should be - as you said - fairly straightforward to implement.

marvinborner:
I've added a bang operator csv generator now. It's not well written and again just a proof of concept, but the parsing of this file should be quite easy and an implementation should really be straightforward now.
My idea is to generate the list on installation so the repository doesn't get cluttered (and you don't need to think about the license of the data). This script could also be executed in the run script with a condition like `if (list doesn't exist) generate_list()`.

ethancedwards8:
Oh man, I just came across this project, I would LOVE for this to get merged. The lack bangs are the only thing stopping me from moving away from DuckDuckGo to another search engine.

marvinborner:
Feel free to add feedback/merge/commit :)

benbusby:
Getting this feature merged is in my near-term plans for the project, I just wanted to finish up the 0.2.1 release beforehand (which is done now, but took me longer than I expected). 

I'll likely have some time this weekend to see what else needs to be done for a basic implementation. I'm thinking it would be nice to also include suggestion support (similar to what DuckDuckGo does now when you type "!"), but depending on the amount of work involved I may separate that effort from this one.

benbusby:
I just committed my changes, so this should be good to go now. @marvinborner (or anyone else reading) if you'd like to test this out before I merge, let me know, otherwise I'll likely merge this in sometime tomorrow. 

I managed to include some basic support for suggesting bang results in the search bar, but the bang json from DDG is absurdly large (something like 12K+ options I think), so I'm not actually sure how useful it is. Maybe at some point this could be improved to filter by most commonly used bangs or something, but at the moment I'm not entirely sure of the best way to do that. For now though it's definitely good enough to merge in and improve as more people use it. 

As a reminder for anyone reading, once this is merged, it'll be immediately available for testing directly on the `develop` branch or using the `beta` tag on Docker Hub.

marvinborner:
Seems alright, thanks!

ethancedwards8:
I'll test it out

benbusby:
I went ahead and merged this so that others can start testing it on the beta tag/develop branch and provide feedback if needed. Thanks again @marvinborner!

PR #106, opened 2020-07-18
Privacy respecting alternatives in results view

### LINKED ISSUE #104: [FEATURE] Option to use privacy-respecting services
It would be nice if we could activate an option to transform all "*twitter.com*" links to "*nitter.net*", "*youtube.com*" to "*invidio.us*", "*maps.google.com*" to "*openstreetmaps.org*" and "*instagram.com*" to "*bibliogram.art*" (or any instance of preference from users).

**Describe which parts of the project this would modify (/app/filter.py)**
I think "filter.py" should be modified (I may be wrong as I'm not very familiarized with the project). Maybe a string.replace("service.bad", "service.good") would do as (at least invidious and nitter) share the exact same links with their respective pairs (eg. twitter.com/Snowden/status/... == nitter.net/Snowden/status/... and same for invidio.us)

**Additional context**
Those services fully respect the privacy from users and allow them to navigate through those privative networks.

### LINKED ISSUE #107: [BUG] "Search only for" doesn't work
**Describe the bug**

Some queries yield "Including results for" output and a link "Search only for" is also provided, but it looks like the link gives the very same results. 

**To Reproduce**

Steps to reproduce the behavior: search for "Searle conscience", look for "Including results for _Searle consciousness_  /  Search only for _Searle conscience_" and click on the latter.

### LINKED ISSUE #108: [BUG] Search query correction not working
**Describe the bug**
When Whoogle automatically searches for another query, slightly different then what you intended to search for, the option to "search instead for ..." does not work.

[dialog box I am referring to](https://user-images.githubusercontent.com/14806962/87979355-2bab4480-ca8f-11ea-8a3d-ed6fc87f5339.png)

**To Reproduce**
1. Search for any query that triggers this behavior, in my case, I searched for "hoogle mail" instead of "google mail", although the issue was not limited to this query
2. Click on "search instead for <query>"
3. It does not change my search results as it should, although the URL does change. Clicking the link after "showing results for <query>" does work

**Deployment Method**
- [ ] Docker
- [ ] `run` executable
- [] pip/pipx
- [x] Other: [describe setup]
Installed manually and run through a systemd service on Ubuntu Server 20.04 as an LXC container on Proxmox

**Version of Whoogle Search**
- [ ] Latest build from [source] (i.e. GitHub, Docker Hub, pip, etc)
- [x] Not sure
I installed it about 1 week ago from source

Void Linux 64 bit, Musl LibC, Firefox 78.0 and IceCat 68.10.0esr browsers.
Debian 64 bit, Firefox 

Thanks! Whoogle has been really great to use!

benbusby:
- Replaces social media links with privacy respecting alternatives (Twitter/YouTube/Instagram -> Nitter/Invidious/Bibliogram).
- Verbatim search now supported
- Option to ignore autocorrect now supported

Open Street Maps support was not included in this effort, since it wasn't as straightforward as the social media alternative functionality. Once more research has gone into how to translate google maps links to osm links, a new PR/issue should be created. 

Closes #104 
Closes #107 
Closes #108

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,4 +1,4 @@
-from app.utils.misc import generate_user_keys
+from app.utils.session_utils import generate_user_keys
 from flask import Flask
 from flask_session import Session
 import os
--- a/app/filter.py
+++ b/app/filter.py
@@ -1,55 +1,10 @@
 from app.request import VALID_PARAMS
-from app.utils.misc import BLACKLIST
-from bs4 import BeautifulSoup
+from app.utils.filter_utils import *
 from bs4.element import ResultSet
 from cryptography.fernet import Fernet
 import re
 import urllib.parse as urlparse
 from urllib.parse import parse_qs
-
-SKIP_ARGS = ['ref_src', 'utm']
-FULL_RES_IMG = '<br/><a href="{}">Full Image</a>'
-GOOG_IMG = '/images/branding/searchlogo/1x/googlelogo'
-LOGO_URL = GOOG_IMG + '_desk'
-BLANK_B64 = '''
-data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAoAAAAKCAQAAAAnOwc2AAAAD0lEQVR42mNkwAIYh7IgAAVVAAuInjI5AAAAAElFTkSuQmCC
-'''
-
-
-def get_first_link(soup):
-    # Replace hrefs with only the intended destination (no "utm" type tags)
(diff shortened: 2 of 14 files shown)


mHatsune:
@benbusby Here's the project profile https://github.com/openstreetmap ;)
Maybe you can link this issue and they may interest to do something.

benbusby:
@mHatsune thanks! I'm actually already familiar with OSM from my day job. I'd prefer to find a solution on my end before bugging them though, that's sort of a last resort imo.

PR #118, opened 2020-08-20
fix pipx run command

cs01:
Add the required `--spec` argument

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ Persistent install:
 
 Sandboxed temporary instance:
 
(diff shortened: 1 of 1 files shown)


benbusby:
Is this a new requirement? When I was testing this originally the `--spec` argument wasn't required, although it's been a while since I originally wrote the documentation.  

Either way, looks good to me.

cs01:
If the package name matches the app name (like `youtube-dl`, for example), the `--spec` argument is not needed. AFAIK this has always been the case.

PR #128, opened 2020-10-01
Fixed annoying browser autocomplete

curlpipe:
image

Hello! Big fan of Whoogle but when I used it, I noticed that Firefox decided it would be a great idea to display it's own auto complete right over the top of Whoogle's. So I made a little change to ensure that this doesn't happen :smile: Hopefully this is all OK. Be sure to let me know if there is anything troubling you about this.

 \- Curlpipe


benbusby (approved):
Looks great, thank you for your contribution!

PR #132, opened 2020-10-06
Adds theming options, nord+solarized themes

apmechev:
- Adds **nord** dark/light theme
- Adds **solarized** dark/light theme
- Adds themes folder in static/css, allows for dark/light theming
- Adds theme option in settings. Automatically parses themes folder to list themes
- Automates the reskin method to use reskin selector in theme.css if possible


app/static/css/themes/nord-light.css
```diff
@@ -0,0 +1,52 @@
+html {
+    background-color: #D8DEE9b3 !important;
+}
+
+body {
+    background-color: #f9fafb !important;
```
benbusby, line 6: How come the `body` element color is different from the `html` one? It should probably just be one or the other. From a personal standpoint, I think `D8DEE9b3` looks better, although you could probably keep it just straight rgb as `D8DEE9` and leave off the alpha.

app/static/css/themes/nord-light.css
```diff
@@ -0,0 +1,52 @@
...
+    color: #3B4252 !important;
+}
+
+#search-bar {
+    color: #4C566A !important;
+    background-color: #ffffff !important;
+}
+
+.search-container {
+    background-color: #f9fafb !important;
```
benbusby, line 41: Does this need to be specified? It should just inherit the color from the body...otherwise if it uses a different color it ends up looking like this:

image

app/static/css/themes/nord-search-light.css
```diff
@@ -0,0 +1,35 @@
...
+    border-bottom: 1px solid #242424;
+}
+
+.autocomplete-items div:hover {
+    background-color: #404040;
+}
+
+.autocomplete-active {
+    background-color: #685e79 !important;
+    color: #f9fafb;
```
benbusby, line 34: The `background-color` here clashes a bit with the default color, and it doesn't look like the `color` is getting applied since it isn't flagged as important.

image

app/static/css/themes/nord-light.css
```diff
@@ -0,0 +1,52 @@
...
+
+footer {
+    background-color: #f9fafb !important;
+    color: #3B4252;
+}
+
+.reskin {
+    background-color: #ffffff;
+    color: #3B4252;
+}
```
benbusby, line 52: I haven't looked through this extensively yet, but this light theme seems to have issues with div colors and/or text color:

image

benbusby (commented):
Looks great so far, I especially like the new nord dark theme! Thanks a bunch for taking care of this, I know it'll make a lot of people happy. I have a few notes from testing it out. 

- The default "whoogle" light theme seems to be missing its own css files, specifically the search suggestion styling is missing.
- A few issues with the nord light theme, see other comments for that.
- I think it might be better if each theme had one single theme file, so that anyone who comes along and wants to add a separate theme can do so with only one added file. Otherwise it could get kind of bloated as more themes get added. 

Let me know your thoughts and I'll re-review whenever this is updated. Thanks again!

apmechev:
Updated the issues with the light mode skins however I'm still thinking how to include dark styling without creating a new .css file.  Maybe templating the .css files?

apmechev:
Added solarized/dark theme as well. Adding a theme should be much easier now. Let me know if the `.css` files make sense; It's my first time really working on a front-end task

benbusby:
@apmechev awesome, thanks for your work on this. I'll do another review soon!

app/filter.py
```diff
@@ -29,13 +31,51 @@ def __getitem__(self, name):
...
+            dict: Key-value dictionary of the properties 
+        """
+        result = {'background-color':'fff'}
+        if self.theme:
+            sheet = cssutils.parseString(self.get_theme_file())
+            if self.dark:
+                selector_name = ".reskin-dark"
+            else:
+                selector_name = ".reskin"
+            reskin_selector = [i for i in sheet if i.selectorText==selector_name]
```
benbusby, line 61: Minor styling nitpick here:
```suggestion
            reskin_selector = [i for i in sheet if i.selectorText == selector_name]
```

app/models/config.py
```diff
@@ -297,6 +300,10 @@ class Config:
         {'name': 'Zimbabwe', 'value': 'countryZW'}
     ]
 
+    THEMES = list(set([theme_name.split('.css')[0]
+                       for theme_name in os.listdir('app/static/css/themes/') 
+                       if "-search.css" not in theme_name]))
+
```
benbusby, line 306: I don't think the cast to `list` is necessary here, since the iterator for themes in the `index.html` template will work fine when provided with a `set`. 

```suggestion
    THEMES = set([theme_name.split('.css')[0]
                       for theme_name in os.listdir('app/static/css/themes/') 
                       if "-search.css" not in theme_name])

```

app/templates/index.html
```diff
@@ -21,14 +21,21 @@
...
         <link rel="search" href="/opensearch.xml" type="application/opensearchdescription+xml" title="Whoogle Search">
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
-        <link rel="stylesheet" href="/static/css/{{ 'search-dark' if config.dark else 'search' }}.css">
         <link rel="stylesheet" href="/static/css/main.css">
-        {% if config.dark %}
-            <link rel="stylesheet" href="/static/css/dark-theme.css"/>
+        {% if config['theme'] %}
+            <link rel="stylesheet" href="/static/css/themes/{{config['theme']}}-search.css" />
+        {% else %}
+            <link rel="stylesheet" href="/static/css/{{ 'dark-' if config['dark'] else '' }}theme.css" />
```
benbusby, line 28: Seems like this could just be simplified to the code below, and remove the if/else block on lines 30-34.

```suggestion
        {% if config['theme'] %}
            <link rel="stylesheet" href="/static/css/themes/{{config['theme']}}.css" />
            <link rel="stylesheet" href="/static/css/themes/{{config['theme']}}-search.css" />
        {% else %}
            <link rel="stylesheet" href="/static/css/{{ 'dark-' if config['dark'] else '' }}theme.css" />
```

app/templates/index.html
```diff
@@ -88,6 +95,16 @@
                             <label for="config-near">Near: </label>
                             <input type="text" name="near" id="config-near" placeholder="City Name">
                         </div>
+                        <div class="config-div">
+                            <label for="config-theme">Theme: </label>
+                            <select name="theme" id="config-theme">
+                                {% for theme in themes %}
+                                <option value="{{ theme }}" {% if theme in config.theme %} selected {% endif %}>
+                                    {{ theme }}
```
benbusby, line 103: Another minor nitpick here, but just to keep styling consistent throughout the config options:
```suggestion
                                    {{ theme.capitalize() }}
```

app/static/css/themes/solarized-search.css
```diff
@@ -0,0 +1,51 @@
...
+    top: 100%;
+    left: 0;
+    right: 0;
+}
+
+.autocomplete-items div {
+    padding: 10px;
+    cursor: pointer;
+    color: #002b36 !important;
+    background-color: #eee8d5 !important;
```
benbusby, line 24: These properties probably shouldn't be flagged as `!important`, since that overrides the properties in `.autocomplete-active`. 

```suggestion
    color: #002b36;
    background-color: #eee8d5;
```

app/static/css/themes/nord-search.css
```diff
@@ -0,0 +1,51 @@
...
+    top: 100%;
+    left: 0;
+    right: 0;
+}
+
+.autocomplete-items div {
+    padding: 10px;
+    cursor: pointer;
+    color: #3B4252 !important;
+    background-color: #ECEFF4 !important;
```
benbusby, line 24: These properties probably shouldn't be flagged as `!important`, since that overrides the properties in `.autocomplete-active`. 

```suggestion
    color: #3B4252;
    background-color: #ECEFF4;
```

benbusby (changes requested):
Just a few small changes needed before merging, thanks again

apmechev:
> Just a few small changes needed before merging, thanks again

All updated!
Thanks for the review/help

benbusby:
I actually just thought of a way to potentially simplify this even more using CSS variables. My idea is that there would be theme stylesheets that declare global theme variables loaded first, and then generic template stylesheets that are loaded second and would refer to the theme variables like so:

```css
html .light {
    background-color: var(--theme-background-light)
    color: var(--theme-foreground-light)
}
html .dark {
    background-color: var(--theme-background-dark)
    color: var(--theme-foreground-dark)
}
```

and the actual theme files would just declare the variables like so:

`nord.css`:
```css
:root {
    --theme-background-light: #eceff4;
    --theme-background-dark: #292e39;

    --theme-foreground-light: ...
    etc.
}
```

`solarized.css`:
```css
:root {
    --theme-background-light: #eee8d5;
    --theme-background-dark: #002b36;

    --theme-foreground-light: ...
    etc.
}
```

Any thoughts on this approach? I realize it would mean a bit more work and refactoring to transition to this method instead, so just let me know if you'd rather just merge this now and worry about that later. I'm fine either way - I think the current implementation is good enough for now. I just thought this could maybe simplify the process even more for future contributors.

apmechev:
Yeah that's a better way to do it for sure. Will update it soon

apmechev:
So, looks like cssutils throws errors when it tries to parse a css file with variables. Refactoring the .css files will be more effort than I though so I suggest merging it now to deploy the feature and I'll open a new PR to refactor the css/python code. Let me know if that sounds good or if I should put more effort in refactoring now

benbusby:
So I've been considering it a bit, and I think the best idea would actually be to do the refactoring now rather than later. As excited as I am for this feature, I want to make sure it gets merged in with the intent of being a "finished" feature rather than with the idea of refactoring and potentially implementing the feature differently later. 

That being said, I'm happy to jump in and handle the refactoring steps since I'm turning it into a hassle at this point. I should be able to take a look at it this week/weekend if that sounds good to you. Or if you'd rather take care of it on your end, that's great too, I just want to make sure I'm not imposing on you if you're too busy or anything.

tborychowski:
Hi guys,
Any update on this PR?
Would be great to add this feature :-)

If css variables are causing issues, IMHO, leave them out for now. 
The UI is very simple and a well designed css theme doesn't really need css variables.
Nor any fancy toggling between dark & light, tbh. 
Not every theme will (or should) have a corresponding light theme.

The most important thing is to add support for custom css (as I can imagine the UI may detract some people, no offense 😉).

Lastly, when it comes to styling & theming - a bigger issue is the html and class naming, as it seems that some sort of a library is used for html generation and the class names are probably not fixed:
Screenshot 2021-02-22 at 14 25 39

which makes styling & theming more difficult/confusing.

benbusby:
If there were updates, they would've been posted here ;)

I disagree about CSS variables being unnecessary. The number of emails and requests I've gotten for X or Y theme would mean that everyone wanting a new theme would just copy/paste boilerplate CSS files over and over again. Having a set of variables in place with a single CSS file would at least allow people to easily add new theme color guides without having to ensure complete parity between the other themes.

That being said, theming is not a priority for me at this point. I don't have tons of free time to spend on the project, so my main focus is functionality and bug fixes, not making it look beautiful (and if anyone is too bothered by the UI to use the project, that's completely fine with me).

The class names are generated on Google's side, so no help there either.

If anyone else wants to finish up this issue in a way that won't create a tsunami of requests for "L33t-amoled-green&blacklololol" theme variants, I'm happy to review their solution and merge it. I'm more inclined to support custom theming rather than individual themes (for the aforementioned reason), but with the obfuscated random class names it seems like it'd be a fairly janky solution.

tborychowski:
I see your points and thank you for clarification 🙂 
I'd love to help but don't know python.

benbusby:
I've worked a bit on adding support for custom CSS theming, merged in #227. The previously discussed approach of implementing CSS variables for tuning color themes to the user's preference was implemented, and the nord and solarized themes created by @apmechev were added [to the wiki here](https://github.com/benbusby/whoogle-search/wiki/User-Contributed-CSS-Themes) -- thank you again for all of your work that you did!

The obfuscated class names will probably always be a problem, but basic page layout styling can still be accomplished (such as wider width results, which seems to be a common request).

PR #139, opened 2020-10-28
Use relative links instead of absolute.

bugbounce:
This allows for hosting under a subpath. For example if you want to host
whoogle at example.com/whoogle, it should work better with a reverse proxy.

Sample nginx reverse proxy config:
```
server {
                server_name whoogle.example.com;
                listen 443 ssl http2;

                location /whoogle/ {
                        proxy_pass http://127.0.0.1:5000/;
                }
```


app/templates/index.html
```diff
@@ -1,37 +1,37 @@
...
+        <link rel="manifest" href="static/img/favicon/manifest.json">
         <meta name="referrer" content="no-referrer">
         <meta name="msapplication-TileColor" content="#ffffff">
-        <meta name="msapplication-TileImage" content="/static/img/favicon/ms-icon-144x144.png">
-        <script type="text/javascript" src="/static/js/autocomplete.js"></script>
-        <script type="text/javascript" src="/static/js/controller.js"></script>
+        <meta name="msapplication-TileImage" content="static/img/favicon/ms-icon-144x144.png">
+        <script type="text/javascript" src="static/js/autocomplete.js"></script>
+        <script type="text/javascript" src="static/js/controller.js"></script>
         <link rel="search" href="/opensearch.xml" type="application/opensearchdescription+xml" title="Whoogle Search">
```
benbusby, line 22: This should probably get updated as well:
```suggestion
        <link rel="search" href="opensearch.xml" type="application/opensearchdescription+xml" title="Whoogle Search">
```

benbusby (changes requested):
Looks good to me, just the one update to the opensearch location. Thanks!

app/templates/index.html
bugbounce: Fixed this. Thanks for pointing out.

PR #157, opened 2020-12-10
Correct Repl.it account requirement

spikecodes:
Repl.it has deprecated the anonymous experience so I have removed the part of the README stating that users can host on Repl.it without an account.


benbusby:
Ah. That's too bad. Thanks for the heads up and for your contribution!

PR #202, opened 2021-02-14
add basic keyboard support

tborychowski:
This adds basic keyboard support on the search results page:
- `ArrowUp` or `k` - focus previous result
- `ArrowDown` or `j` - focus next result
- `/` focus search box

Because this just focuses links in the search result items hitting `Enter` (or ctrl/cmd + Enter) when an item is focused - will open the search result (because the link is already focused and this is browser's default behaviour).

### DIFF
--- a/app/static/css/header.css
+++ b/app/static/css/header.css
@@ -60,3 +60,9 @@ header {
     margin: 15px;
     display: block;
 }
+
+
+#main>div:focus-within {
+    border-radius: 8px;
+    box-shadow: 0 0 6px 1px #2375e8;
+}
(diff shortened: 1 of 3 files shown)


benbusby:
Thanks for implementing this, I'm definitely a fan. 

One issue though -- the `j` and `k` navigation is still active when the search input is focused, so typing either letter triggers the navigation action rather than modifying the query with the new letter. Once that's been fixed I think this should be good to go.

tborychowski:
Ha! Sorry about that! :-)
That's the one-man testing 😛
Will fix later today.

tborychowski:
@benbusby should be good now. Thanks!

benbusby (approved):
Awesome, thanks! Looks great.

PR #208, opened 2021-02-22
Add Docker Security Best Practices to docker-compose file.

pred2k:
This adds several security improvement to the docker-compose file to protect the host system.

- Don't run container as root user. Container processes shouldn't have root priviliges to the Kernel on host system.
- Run container with a read-only main volume. Container are immutable instances. Use tmpfs if the app need to write temporary files.
- Limit resources like RAM and PIDs.

From **CIS Docker Community Edition Benchmark** and **[OWASP Docker Security Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html)**.

I couldn't successfully test a search connection over tor. This didn't even work with a original `benbusby/whoogle-search` image.  
I misuse the `debian-tor` user (uid 102) to get the `tor` and the `python` process running in the image.

Additionaly i would like to discuss two things:
- It would be better to create separate user in the `Dockerfile` (and set it with `USER ...`) to run `python3 -um app ...`. This allows more specific permission grants for the python app.
- The `tor` process should be moved a seperate container and started as a second service via the docker-compose file. This could also make the tor connection more optional.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,9 +1,26 @@
-version: "3"
+# cant use mem_limit in a 3.x docker-compose file in non swarm mode
+# see https://github.com/docker/compose/issues/4513
+version: "2.4"
 
 services:
   whoogle-search:
     image: benbusby/whoogle-search
     container_name: whoogle-search
+    restart: on-failure:5
+    pids_limit: 50
+    mem_limit: 256mb
+    memswap_limit: 256mb
+    # user debian-tor from tor package
+    user: '102'
+    security_opt:
+      - no-new-privileges
+    cap_drop:
+      - ALL
+    read_only: true
+    tmpfs:
+      - /config/:size=10M,uid=102,gid=102,mode=1700
+      - /var/lib/tor/:size=10M,uid=102,gid=102,mode=1700
+      - /run/tor/:size=1M,uid=102,gid=102,mode=1700
     #environment: # Uncomment to configure environment variables
       # Basic auth configuration, uncomment to enable
       #- WHOOGLE_USER=<auth username>


benbusby:
> I couldn't successfully test a search connection over tor. This didn't even work with a original benbusby/whoogle-search image.

That's strange, as the Heroku instances (which are fairly barebones) can start the tor process fine. 

> The tor process should be moved a seperate container and started as a second service via the docker-compose file.

I agree in theory, but I believe this will break the structure that Heroku enforces, where all application processes must be started by a single Dockerfile. I could be wrong, might be worth looking into some more. 

>This could also make the tor connection more optional.

The connection already is optional. If the connection fails to start, the app doesn't care, it just labels the service as "Unavailable". 

> It would be better to create separate user in the Dockerfile (and set it with USER ...) to run python3 -um app .... This allows more specific permission grants for the python app.

No disagreements here, and shouldn't introduce any problems. Feel free to update the PR with those changes, and we can get it merged in soon.

pred2k:
> > I couldn't successfully test a search connection over tor. This didn't even work with a original benbusby/whoogle-search image.
> 
> That's strange, as the Heroku instances (which are fairly barebones) can start the tor process fine.

In the container image of this PR i can see more bootstrap logs from Tor, like building the Tor circuit and the line `[notice] New control connection opened from 127.0.0.1.`. Which i dont see the upstream image. Maybe they are hidden.
It's sad that i cant verify the search via tor.

> > The tor process should be moved a seperate container and started as a second service via the docker-compose file.
> 
> I agree in theory, but I believe this will break the structure that Heroku enforces, where all application processes must be started by a single Dockerfile. I could be wrong, might be worth looking into some more.

I unterstand heroku is a important way to depoy whoogle. I thought about the idea of an whoogle-image tag without tor included.

> > It would be better to create separate user in the Dockerfile (and set it with USER ...) to run python3 -um app .... This allows more specific permission grants for the python app.
> 
> No disagreements here, and shouldn't introduce any problems. Feel free to update the PR with those changes, and we can get it merged in soon.

I'm not sure about the benefits to add another user for whoogle besides `debian-tor`, if they still run in the same container enviroment.

benbusby (approved):
Sorry for the delay, been super swamped lately. This all looks good.

At some point in the future I think it'd be worth revisiting whether or not Tor is actually necessary in the project. With how often Google blocks exit nodes, it only ever seems to work for me maybe 1/10 times, and that's even with automatic identity resetting. I don't mind keeping it for now since it isn't hurting anything, but I'm curious how much value it's actually providing.

PR #210, opened 2021-02-25
Use multi-stage container build

pred2k:
This only add necessary packages and files from the repo to reduce the image size.

The image size drops from 566MB to 269MB.


benbusby (approved):
Looks great, thanks!

PR #231, opened 2021-03-20
Fix manifest.json not loading

shimuldn:
The error on the console fix
`Refused to load manifest from 'https://xxxxx.xxx/static/img/favicon/manifest.json' because it violates the following Content Security Policy directive: "default-src 'none'". Note that 'manifest-src' was not explicitly set, so 'default-src' is used as a fallback.`


benbusby (approved):
Looks good, thanks!

PR #234, opened 2021-03-21
Configuring PWA for Mobile Browser

shimuldn:
Finxing PWA issue for Mobile Phones.
icon loading issue.

Before: 
Console error

Application error
No logo

Before

After this fix:

Install on firefox.
Install    Settings menu

We got logo!
Add to home screen now

I did checked on few browser and iPhone.
Looks good to me.

### DIFF
--- a/app/static/img/favicon/manifest.json
+++ b/app/static/img/favicon/manifest.json
@@ -3,7 +3,6 @@
  "short_name": "Whoogle",
  "display": "fullscreen",
  "scope": "/",
- "start_url": "$APP_ROOT/templates/index.html",
  "icons": [
   {
    "src": "android-icon-36x36.png",
--- a/app/static/img/favicon/manifest.json
+++ b/app/static/img/favicon/manifest.json
@@ -1,38 +1,42 @@
 {
- "name": "App",
+ "name": "Whoogle Search",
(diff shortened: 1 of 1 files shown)


app/static/img/favicon/manifest.json
```diff
@@ -1,41 +1,45 @@
 {
- "name": "App",
+ "name": "Whoogle Search",
+ "short_name": "Whoogle",
+ "display": "fullscreen",
+ "scope": "/",
+ "start_url": "$APP_ROOT/templates/index.html",
```
benbusby, line 6: Is this necessary? I don't think `APP_ROOT` is actually being used here, and it looks like everything else should work without a `start_url` property.

```suggestion
```

benbusby (commented):
One small question/suggestion, otherwise looks good.

app/static/img/favicon/manifest.json
shimuldn: I was looking with the sample https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Installable_PWAs
Yes, Work with `start_url` as well.

benbusby (approved):
Thanks, looks good.

PR #237, opened 2021-03-21
Setting up alternative social media site using .env

shimuldn:
The goal to support the .env file for setting social media site, dark mode, Language, City, Near and some other Settings at the time of installations so the same .env file can be used in case of reinstallation.

Currently .env working for social media site `Manual` installation.


shimuldn:
.env not reading by docker currently.

shimuldn:
@benbusby any suggestion?

benbusby:
I haven't looked at this too much yet, but it looks like `.env`isn't getting copied over into the container.

Overall though I think this will probably need a modified approach to make sure that user provided environment variables aren't overwritten by the `.env` file (i.e. having a blank `.env` file by default) and then conditionally loading values from the file if a separate flag is set (like `WHOOGLE_DOTENV` or something). From what it looks like now, setting `WHOOGLE_ALT_TW` or any of the other environment variables manually would be superseded by the values you have in the `.env` file. Ideally these would also use the same naming scheme to avoid any confusion. 

Haven't had time to test this out or think much about it yet though, I'll take a deeper look once I have some more time. Depending on how much needs to be changed I may just add my proposed solution directly to your branch, but I'll let you know.

benbusby (approved):
I made a few changes. Environment variables are now set in `whoogle.env` rather than `.env` so that it's more obvious to the average first-time user. Values in the file are loaded if the `WHOOGLE_DOTENV` value is set during deployment. So as long as `WHOOGLE_DOTENV=1` is set before running, the values from `whoogle.env` will be loaded and override any other default environment variables.

shimuldn:
Thank you.

benbusby:
Yes, you can run it like this `WHOOGLE_DOTENV=1 ./run`

PR #264, opened 2021-04-05
Add wget to container build otherwise the healthcheck won't work.

Suika:
You kinda need a wget binary in the container if you are gonna use wget in the healthcheck.


benbusby (approved):
Huh. I thought this was already included by default in the container. Doesn't hurt to include either way though.

PR #268, opened 2021-04-06
View image functionality

jacr13:
Adds the view image functionality using mobile user agent to get real image URLs.


benbusby:
Nice, this works really well from what I can tell after a bit of preliminary testing. I'd like to review this a bit more in depth since it's a somewhat large change, and I'm hoping to have some time this week to do so. Possibly even today, but I'll let you know. Either way thanks for looking into this -- I'm excited to get this merged in.

jacr13:
Thanks, I think maybe I should change the behavior when we click the thumbnail so it redirects to the full-res image instead of the article?!

app/templates/imageresults.html
```diff
@@ -0,0 +1,119 @@
...
+                      {{ results[(i*4)+j].domain }}
+                    </span>
+                    </span>
+                  </div>
+                  </a>
+                  <a href="{{ results[(i*4)+j].img_url }}">
+                    <div class="Tor4Ec">
+                      <span class="qXLe6d F9iS2e">
+                        <span class="fYyStc">
+                          View image
```
benbusby, line 92: Could you pass this in as a template string? Something like `view_label="View Image"` and then adding here as `{{ view_label }}`. It'd be nice to reduce the number of hardcoded strings in the templates in order to better support localization in the future.

app/templates/index.html
```diff
@@ -118,6 +118,12 @@
                             <label for="config-new-tab">Open Links in New Tab: </label>
                             <input type="checkbox" name="new_tab" id="config-new-tab" {{ 'checked' if config.new_tab else '' }}>
                         </div>
+                        <div class="config-div config-div-view-image">
+                          <label for="config-view-image">View image: </label>
+                          <input type="checkbox" name="view_image" id="config-view-image" {{ 'checked' if config.view_image else '' }}>
+                          <div><span class="info-text"> — Adds the View Image option.
+                            The thumbnails are blurry because mobile thumbnails are smaller.</span></div>
```
benbusby, line 125: Just some improved clarification here.

```suggestion
                          <div><span class="info-text"> — Adds the "View Image" option to image search results.
                            Note: This will cause image result thumbnails to be lower resolution.</span></div> 
```

benbusby (changes requested):
Overall nice work! A few minor changes needed, nothing major.

I do wish there was a simpler way to include support for viewing full size images, but this seems to be the best way that doesn't involve making multiple requests to both desktop and mobile image search results. 

You'll also need to pull in the changes from `develop` that introduce the new approach for setting boolean config values, but that should be straightforward.

app/templates/imageresults.html
```diff
@@ -0,0 +1,119 @@
...
...XLRf{vertical-align:top}.MhzMZd{border:0;vertical-align:middle;font-size:14px;height:40px;padding:0;width:100%;padding-left:16px}.xB0fq{height:40px;border:none;font-size:14px;background-color:#4285f4;color:#fff;padding:0 16px;margin:0;vertical-align:top;cursor:pointer}.xB0fq:focus{border:1px solid #000}.M7pB2{border:thin solid #dadce0;margin:0 0 3px 0;font-size:13px;font-weight:500;height:40px}.euZec{width:100%;height:40px;text-align:center;border-spacing:0}table.euZec td{padding:0;width:25%}.QIqI7{display:inline-block;padding-top:4px;font-weight:bold;color:#4285f4}.EY24We{border-bottom:2px solid #4285f4}.CsQyDc{display:inline-block;color:#70757a}.TuS8Ad{font-size:14px}.HddGcc{padding:8px;color:#70757a}.dzp8ae{font-weight:bold;color:#3c4043}.rEM8G{color:#70757a}.bookcf{table-layout:fixed;width:100%;border-spacing:0}.InWNIe{text-align:center}.uZgmoc{border:thin solid #dadce0;color:#70757a;font-size:14px;text-align:center;table-layout:fixed;width:100%}.frGj1b{display:block;padding:12px 0 12px 0;width:100%}.BnJWBc{text-align:center;padding:6px 0 13px 0;height:35px}.e3goi{vertical-align:top;padding:0;height:180px}.GpQGbf{margin:auto;border-collapse:collapse;border-spacing:0;width:100%}
```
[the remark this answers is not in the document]
jacr13: fix "Basically the .cvifge class needs width: 100%; in order to expand the search input to fit the form width."

app/templates/imageresults.html
jacr13: Yes will push the fix

jacr13:
I have pushed the changes, thanks!

benbusby (approved):
Great, looks good. Thanks again for taking care of this!

PR #269, opened 2021-04-06
added public instance

igosad:


Hunam6:
image

igosad:
> image

The URL was `https://www.whoogle.tech/` and not `https://whoogle.tech/`.
 I changed then subdomain so it can be more clear .

benbusby (approved):
Looks good, thanks!

I'm planning on expanding the public instances list into a table with more information in the near future. Just to confirm, it looks like this instance is hosted in the US?

igosad:
>this instance is hosted in the US?
Yes it is hosted on Google cloud servers + Cloudflare, and it will change IPs every 24 hours.

PR #273, opened 2021-04-09
Update README.md

ForyxCraft:
Added another public Whoogle instance


benbusby (approved):
Thanks! What country is this hosted in?

PR #278, opened 2021-04-10
always open non-result links in same window

kaneorotar:
The current behavior of turning on "Open Links in New Tab" makes all links open in new window, including "Videos", "News", "Images", and even the "Next>" button!

This PR sets target attribute to "_self" for non-result links so that they will not be opened in a new window.

### DIFF
--- a/app/filter.py
+++ b/app/filter.py
@@ -232,6 +232,7 @@ class Filter:
                 param_val = query_params[param][0]
                 new_search += '&' + param + '=' + param_val
             link['href'] = new_search
+            link['target'] = '_self'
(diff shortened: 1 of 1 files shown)


jacr13:
I completely agree with you, the new tab option should only open external links in a new tab.

However, the fix you propose doesn't solve completely the problem, e.g. the links starting with "/?safe". 

A couple of days ago, I pushed a fix to this to another PR so I removed it, but you can find the fix 
[here](https://github.com/benbusby/whoogle-search/pull/268/commits/ba296f056a2fc264eb3e6f52a813d553750bb105).

PR #295, opened 2021-04-22
Option to disable changing configuration

notangelmario:
Added a simple option to disable changing of configuration. I don't think it's something necessary but it's a nice thing to have. I guess public instances could have something like this enabled. Testing is not implemented yet.

### DIFF
--- a/app.json
+++ b/app.json
@@ -75,6 +75,12 @@
         "value": "",
         "required": false
     },
+    "WHOOGLE_CONFIG_DISABLE_CHANGE": {
+        "description": "[CONFIG] Disable ability for client to change config (set to 1 or leave blank)",
+        "value": "",
+        "required": false
+    },
+
     "WHOOGLE_CONFIG_DARK": {
         "description": "[CONFIG] Enable dark mode (set to 1 or leave blank)",
         "value": "",
--- a/app/routes.py
+++ b/app/routes.py
@@ -129,6 +129,7 @@ def index():
                            logo=render_template(
                                'logo.html',
                                dark=g.user_config.dark),
+                           config_disabled=os.getenv('WHOOGLE_CONFIG_DISABLE_CHANGE', ''),
                            config=g.user_config,
                            tor_available=int(os.environ.get('TOR_AVAILABLE')),
                            version_number=app.config['VERSION_NUMBER'])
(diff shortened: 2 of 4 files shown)


app.json
```diff
@@ -75,6 +75,12 @@
         "value": "",
         "required": false
     },
+    "WHOOGLE_CONFIG_DISABLE_CHANGE": {
```
benbusby, line 78: I think just `WHOOGLE_CONFIG_DISABLE` would be a better name, the fact that it disables changes specifically is implied (just my opinion, let me know if you disagree). 

If you think it should include the `CHANGE` qualifier, my only nitpick would be to update it to `WHOOGLE_CONFIG_DISABLE_CHANGES`.

app/routes.py
```diff
@@ -129,6 +129,7 @@ def index():
                            logo=render_template(
                                'logo.html',
                                dark=g.user_config.dark),
+                           config_disabled=os.getenv('WHOOGLE_CONFIG_DISABLE_CHANGE', ''),
```
benbusby, line 132: I'd prefer if this was an app level config setting (i.e. in `app/__init__.py`). Then access it via `app.config['CONFIG_DISABLED']` or whatever, rather than reading from the environment variable directly.

app/routes.py
```diff
@@ -239,7 +240,7 @@ def search():
 def config():
     if request.method == 'GET':
         return json.dumps(g.user_config.__dict__)
-    elif request.method == 'PUT':
+    elif request.method == 'PUT' and os.getenv('WHOOGLE_CONFIG_DISABLE_CHANGE', '') == '':
```
benbusby, line 243: Would be better as

```python
elif request.method == 'PUT' and not  <...>  # config disabled condition
```

rather than just comparing to an empty string. Also the "disabled config" setting should be moved up to the top of the function definition as a single variable and reused here and below in the `elif` to minimize the amount of work needed for any future refactoring of the property.

app/routes.py
```diff
@@ -250,7 +251,7 @@ def config():
             return json.dumps(session['config'])
         else:
             return json.dumps({})
-    else:
+    elif os.getenv('WHOOGLE_CONFIG_DISABLE_CHANGE', '') == '':
```
benbusby, line 254: See comment above (using `not <variable>` for the check)

app/routes.py
```diff
@@ -270,6 +271,8 @@ def config():
 
         session['config'] = config_data
         return redirect(config_data['url'])
+    else:
+        return json.dumps({})
```
benbusby, line 275: We probably want to just redirect back to home in this case, right? The default behavior of updating the config (in a non-disabled environment) is to redirect back to home, so I think it should be the same in this case.

app/templates/index.html
```diff
@@ -56,111 +56,113 @@
...
-                        <div class="config-div">
-                            <input type="submit" id="config-load" value="Load">&nbsp;
-                            <input type="submit" id="config-submit" value="Apply">&nbsp;
-                            <input type="submit" id="config-save" value="Save As...">
-                        </div>
-                    </form>
-                </div>
-            </div>
-        </div>
+	    {% if config_disabled == '' %}
```
benbusby, line 59:
```suggestion
	    {% if not config_disabled %}
```

benbusby (changes requested):
Looks good, thanks for implementing -- I know quite a few people are going to appreciate having this feature around. 

A few fairly minor changes needed, plus it would be nice to have a test to ensure that when this variable is set, the API requests with the config don't actually modify it. Let me know if you want any help with that aspect.

benbusby (approved):

Issue #311, opened 2021-05-14
[BUG] EU consent enforced, no more pending period


Suika:
**Describe the bug**
Time to accept the overlords has come. #243 is now useless, there is no more waiting period.
If you block js, you are instantly sent to the consent page. Whatever you do, won't change it.
If you have the JS enabled, there is some background worker magic happening, but also a pain to extract.
So the only way now, is to set the cookie to something that is generated.

Example url is `https://consent.google.de/ml?continue=https://www.google.de/&gl=DE&pc=shp&hl=de&src=1`, any change outside `continue` will error out the consent page.
`YES+cb.20210509-17-p0.de+F+678` is an example with everything set to no.
`YES+cb.20210509-17-p0.de+FX+678` is at least one or all parts accepted
Can be shortened to `YES+1`, but probably shouldn't.

https://github.com/benbusby/whoogle-search/blob/develop/app/request.py#L228 needs to be changed.

¯\\_(ツ)_/¯ It's only a matter of time untill you have to scan your subdermal chip to be able to search in the future.

**To Reproduce**
1. Be in EU
2. Use whoogle
2.1. Or deactiavate js, delete cookies and access google 
3. See error

**Deployment Method**
- [ ] Heroku (one-click deploy)
- [X] Docker
- [ ] `run` executable
- [ ] pip/pipx
- [ ] Other: [describe setup]

**Version of Whoogle Search**
- [X] Latest build from [source] (i.e. GitHub, Docker Hub, pip, etc)
- [X] Version [v0.4.1]
- [ ] Not sure

**Desktop (please complete the following information):**
 - doesn't matter

**Smartphone (please complete the following information):**
 - doesn't matter

**Additional context**
#243 was a way to circumvent the agreement, but it no longer works.

ra-pha:
Same problem here on 0.4.1

Further problem is that clicking "I agree" doesn't do anything.

If I want to adjust consent settings then a new tab opens with an error 400.

Akruidenberg:
Same here

bergfex71:
Confirmed.

joonas-fi:
I must admit the bug description is confusing to me (waiting period? JavaScript disabled?), but if it boils down to "Whoogle doesn't work, because it is showing Google cookie consent dialog whose 'I agree' button doesn't work", then yes, I am experiencing the same issue.

JorelAli:
I guess I'll join the list of people dealing with this issue as well!

[...]

PR #320, opened 2021-05-17
Fix EU consent bug

bruvv:
Fix: https://github.com/benbusby/whoogle-search/issues/311


benbusby (approved):
Thanks! I made a small change to format the cookie with the current datetime, as I'm skeptical that the date used in the initial commit will work in the long term.

foxkumar:
@benbusby can you share what steps need to be taken from a layman POV to resolve this.

PR #330, opened 2021-05-27
Add instructions to deploy to Fly.io

jeromegn:
Fly.io is a geo-distributed general-purpose app hosting service. By default it will deploy your app to a location near you, making it a good choice for self-hosted applications such as Whoogle.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,9 +17,10 @@ Contents
     2. [Repl.it](#b-replit)
     3. [pipx](#c-pipx)
     4. [pip](#d-pip)
-    5. [Manual](#e-manual)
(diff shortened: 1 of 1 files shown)


benbusby (changes requested):
Looks great, thanks!

One minor nitpick though -- would you mind moving the Fly.io section to item 3 in the readme (right after Repl.it)? I think it'd make sense to be grouped with the other account-based deployment methods. I know the way I have everything lettered/numbered can be a little tedious to update by hand, so I'm happy to do it myself if you'd like.

benbusby (approved):

PR #337, opened 2021-06-01
Fixed German name

Myzel394:
Translators often translate "German" to "Deutsche", but in this context you use "Deutsch". "Deutsche" is used as an adjective.


benbusby:
Merged in #339

PR #340, opened 2021-06-01
Make search input autofocused

Myzel394:


benbusby (commented):
Autofocus is already enabled for this input: https://github.com/benbusby/whoogle-search/blob/cbcbc26d9b7766ebc52356a4557992828ddb79b2/app/templates/index.html#L51

PR #352, opened 2021-06-09
Fixes redirection to the results page when using containers in firefox.

jacr13:
Fixes redirection to the results page when using containers in Firefox.

The current behavior doesn't take the search words when we type them in the search bar of another container, and instead it redirects to the home page.


benbusby (commented):
This would expose search terms in logs for users using POST requests (the default) for searches and suggestions. The addition you made is how the opensearch template formats the search URL when the `GET Requests Only` config option is enabled, but it seems like Firefox aggressively caches the opensearch template and can be tricky to get it to work with containers.

The way I've gotten this to work with containers is with the following steps:
1. Remove any existing Whoogle search engines from Firefox settings
2. Enable `GET Requests Only` in Whoogle config
3. Clear Firefox cache
4. Restart Firefox
5. Navigate to Whoogle instance and re-add the engine

With these steps I've successfully been able to add the engine and use it with Firefox Containers, but it's a fairly cumbersome process. Unfortunately it's a limitation of Firefox Containers not allowing `POST` request follow through (which I believe there's an issue open for on their end) and the caching of the opensearch template on page load (before the `GET Requests Only` config can be enabled and reflected in the template).

I'll update the README with instructions for enabling the engine when using Firefox Containers and close this PR once that's done, but thanks for bringing this to my attention.

PR #378, opened 2021-07-16
Make replit install all requirements first

KokoTheBest:
This should install all requirements from requirements.txt. It makes this a one click experience, without the user having to run `pip install -r requirements.txt` and then tap the run button. I myself had to first run that command in my repl, so I have made this change so others don't have to do the same.

repl.it also runs on linux based systems, so `&&` is the correct bash syntax.

### DIFF
--- a/.replit
+++ b/.replit
@@ -1,3 +1,3 @@
 language = "python3"
 run = "./run"
-onBoot = "./run"
+onBoot = "pip install -r requirements.txt && ./run"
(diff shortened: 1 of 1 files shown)


benbusby:
Thanks for updating. Originally repl.it automatically detected requirements.txt and installed all dependencies before launching, so I guess they're changing some things on their end.

benbusby (approved):

PR #405, opened 2021-08-27
Fix some portuguese translations

davidfrickert:
Noticed some portuguese translations weren't great so this PR helps a bit on that!


benbusby (approved):
Thanks!

PR #429, opened 2021-09-15
Take the conventional name instead.

gugod:
Not sure how the list in countries.json are generated or whether that should be manually edited -- if somewhere else needs to be edited together please let me know.


benbusby (approved):

benbusby:
It's all currently done manually, so this is correct. Thanks!

PR #430, opened 2021-09-16
:sparkles: update readme :sparkles:

TarunavBA:


benbusby:
Hmmm, I'm personally not a fan of bold text unless it pertains to something critical that would be easily skimmed over. Since the info being bolded is in the first sentence, I don't think anything there would be easily missed, so I think it should be left alone. 

I appreciate the contribution though! Just a personal preference of mine.

PR #433, opened 2021-09-18
:sparkles: update readme

TarunavBA:


TarunavBA:
Added `whoogle-search`

PR #438, opened 2021-09-22
Made Capitalised Bang Search to Work

BlissOwl:


app/utils/bangs.py
```diff
@@ -51,7 +51,8 @@ def resolve_bang(query: str, bangs_dict: dict) -> str:
              wasn't a match or didn't contain a bang operator
 
     """
-    split_query = query.split(' ')
+    splited_query = query.split(' ')
+    split_query = [split_query_lower.lower() for split_query_lower in splited_query]
```
benbusby, line 55: This actually causes a problem where the bang search isn't removed from the query, since the query itself is still case sensitive after resolving which bang was used. Can be simplified to:

```suggestion
    # Ensure bang search is case insensitive
    query = query.lower()
    split_query = query.split(' ')
```

benbusby (changes requested):
Looks good, just one change needed.

benbusby (approved):

Issue #453, opened 2021-10-07
[QUESTION] Does whoogle really need to be run as root?


astier:
Hi. First of all great program. I see that in `whoogle.service` root is set as the user. Does whoogle really need root-privileges? Searx for example creates a new searx-user with limited privileges. Running as root just doesn't seem very secure.

Albonycal:
nope, I don't think whoogle needs root privileges

PR #461, opened 2021-10-14
Use DynamicUser in whoogle.service #453

astier:
Quote from https://www.freedesktop.org/software/systemd/man/systemd.exec.html

> DynamicUser= takes a boolean parameter. If set, a UNIX user and group
pair is allocated dynamically when the unit is started, and released as
soon as it is stopped. The user and group will not be added to
/etc/passwd or /etc/group, but are managed transiently during runtime.

`DynamicUser` creates dynamically a new user `whoogle` with limited
priviliges which are handled by systemd every time whoogle.service is
started. I think this is more secure and easier rather than letting the
admin deal with the user and its priviliges under which whoogle should
run. Its certainly a big security improvement compared to the root-user
which was used to run whoogle not that long ago.

I changed my `whoogle.service` file accordingly and running `ps u -U
whoogle` shows me that systemd automatically runs whoogle with the user
`whoogle` which systemd created automatically because I never created
the user myself.

```
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
whoogle     6299  0.0  0.7 363884 58808 ?        Ssl  20:14   0:01 /opt/whoogle-search/venv/bin/python3 -um app --host 0.0.0.0 --port 5000
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ Description=Whoogle
 # Load values from dotenv only
 #Environment=WHOOGLE_DOTENV=1
 Type=simple
-User=<username>
+DynamicUser=true
 WorkingDirectory=<whoogle_directory>
 ExecStart=<whoogle_directory>/venv/bin/python3 -um app --host 0.0.0.0 --port 5000
 ExecReload=/bin/kill -HUP $MAINPID


astier:
Forget it. With this setting it is not possible for whoogle to apply configuration via the configuration-menu. It aborts with the following error:

```
Oct 14 22:51:38 e490 whoogle[2520]:   File "/opt/whoogle-search/app/routes.py", line 303, in config
Oct 14 22:51:38 e490 whoogle[2520]:     open(app.config['DEFAULT_CONFIG'], 'w').write(
Oct 14 22:51:38 e490 whoogle[2520]: OSError: [Errno 30] Read-only file system: '/opt/whoogle-search/app/static/config/config.json'
```

The issue is that although `DynamicUser=true` would make whoogle generally more secure because of its restrictive nature, only a few directories are writable to whoogle when this setting is enabled. As far as I understand this are `/tmp`, `/var/tmp`, `/var/lib/`, `/var/cache/`, `/var/log/`. Directories under this directories can be made writable with `StateDirectory=`, `CacheDirectory=`, `LogsDirectory=`. As far as I understand `/opt/whoogle-search/app/static/config/` can not be made writable to whoogle with `DynamicUser=true` enabled.

This means code-changes would need to be made such that whoogle would store its configuration only at the allowed places.

Issue #305, opened 2021-05-06
[FEATURE] Update available warning


justsevrage:
**Describe the feature you'd like to see added**
Was thinking about adding some kind of way to notify admins that there is an update available. Maybe display something in the console, maybe add a little notice at the footer. Oh yeah, speaking of the footer, it is not centered properly. Might make a pull request later if no one does.

Issue #383, opened 2021-08-05
[FEATURE] Multiple Proxies


kitchenutensils778:
Is is possible to add multiple proxies and select the least latency proxy in real time?

vacom13:
@benbusby I would like to give this a try. However, I would need guidance on it 😅. I don't know, maybe some resources to refer too?

vacom13:
@benbusby I will take it up

benbusby:
Thanks @vacom13, and sorry I missed your message from back in December! I think for an initial implementation you could just add support for multiple proxies, and retry requests using a different proxy (if multiple are configured) if one times out. Selecting the least latency proxy can be added later as an improvement.

vacom13:
@benbusby no problem. And yes I will look into it

vacom13:
@benbusby Hey. I have actually been really busy with family and college work. I did think up a way to implement this I suppose. Correct me if I am wrong. For now I should probably just get a list of working proxies from a site and give the user a checkbox in the configs to select if he wants to use the proxies right? And them I need to make sure that the search results come within the given timeframe. After that I can give the user the ability to select the latency right?

vacom13:
@benbusby there is this python library `free-proxies`. It basically scrapes for proxies on `https://www.sslproxies.org/` and gives a string for working proxies. I could use that and incase it times out, I could try to get another proxy. As it gets the proxy in real time, I suppose there shouldnt be a problem as according to the documentation, it checks whether the proxy is working.

vacom13:
Or I could just scrape the list myself for the first 5 proxies and then check those out?

benbusby:
Hey @vacom13, I think that's actually a good idea, but potentially a bit out of scope for this issue. I think all this issue should really support is multiple user specified proxies. So if a user has access to multiple, they can specify them as a comma separated string (or something along those lines). 

So currently `WHOOGLE_PROXY_LOC` only accepts one `IP:PORT` string, but could be updated to have multiple and cycle through them if one of them returns an error response code. If a user just specified the proxy locations as a comma separated string such as `IP:4000,IP:4001`, then in the request module we could do something like:

```py
proxy_paths = os.environ.get('WHOOGLE_PROXY_LOC', '').split(',')
if proxy_paths:
    # ...
    for path in proxy_paths:
        # Validate a 200 response and no captcha from search URL
```

I think your idea could be an entirely separate issue, but I'd like to personally look into the `free-proxies` library a bit first.

[...]

Issue #301, opened 2021-04-29
[FEATURE] Add "hours" ranges to Place result


FoxxMD:
**Describe the feature you'd like to see added**
When whoogle returns a specific place as the result it displays open/close status and what time it opens/closes.
When google returns a specific place as the result it displays the same thing but also has dropdown that expands to show all hours of operations.

Please add this dropdown/expand functionality to display all hours of operation in the whoogle view.

**Additional context**
![image](https://user-images.githubusercontent.com/4663766/116567743-e6495c80-a8d5-11eb-84bb-dd1aab7e6896.png)

benbusby:
Oh wow, the flying biscuit! I've only been there once when I visited Atlanta, but I remember it being really good. I even bought a T-shirt. 

Anyways, this might be sort of a tricky thing to add, since Google is doing something behind the scenes to generate that hours list that the non-javascript version of the result page is apparently not privy to. From a little bit of testing it looks like appending "hours" to the search typically returns a more detailed breakdown of the business hours, but the response body of that search isn't lightweight or reliable enough to just be appended to the general business results.

I'll think this over a bit and hopefully get back to you soon on what I come up with.

FoxxMD:
Understandable, thanks for taking a look

Issue #129, opened 2020-10-05
[FEATURE] Currency conversion


Aikatsui:
1. Time
Whoogle display time in 24 format. search: pakistan time, google display it in AM/PM format

2. Currency conversation
This feature not available on whoogle. search: eur to usd
(You can attach cool component for that)

3. Weather
This feature not available on whoogle. search: pakistan weather or certain city weather
(You can ignore if difficult to implement)

4. Calculator
This feature not available for whoogle. search: calculator

(You may need to integrate other services if difficult to take from google)

So far i only expect solution for 1 and 2.

accountForIssues:
I think 2-4 are not possible to implement as Google loads this via javascript in the browser whereas Whoogle scrapes and parses the results from a no-javascript version of Google search results.

@benbusby I thought it was possible to view (basic) results to such queries (1-3) using Whoogle. I remember it worked some versions ago (maybe I misremember). Currently on the latest version, none seem to work. Using the same query on Google no-js page (`gbv=1`), they show basic results for the query.

Example queries;
- usd to gbp
- weather nyc
- time london
- 15*88
- dog in polish

Usually, the first `div` in the search results is the basic answer response to such queries. I know Google makes it difficult by randomizing the class names often but is it possible for Whoogle to detect/scrape these and show it ? Currently, it feels like they are being removed as all other results remain.

IniUe:
This may make delay to implement this feature but it's not a problem, if we can go to next level than to google search then it'll be better too while privacy friendly. so i like to add a comment (enhancement request) about 2 and 3. I guess there're many free data providing services sites and APIs. (If that sounds like forcing, there's no thing like you should implement, that's only suggestions to make this project better :smile:) 

**Currency conversation**
e.g. the search engine get data from morningstar and coinbase for cryptocurrency
https://developers.coinbase.com (Retrieve real-time or historical price information) that's only an example; not as suggestion to use it or particular api. i don't know about its usability and on limited environments etc or possible to scrape from sites.

i'm interest to see whoogle usable to view day/week/years currency statistics charts, contents from economic calendar and other details useful for particular field persons. it'll be amazing for other whoogle users.

**Weather**
I assume this feature hasn't worked when @Aikatsui opened this issue but now it seems work to show weather in specific country or city from weather.com

![image](https://user-images.githubusercontent.com/81799342/122598831-8da56d00-d097-11eb-8999-cbafb022e43c.png)
[...]

PR #517, opened 2021-11-01
Add Update Available warning

vacom13:
Update available now shows up in the footer if "version_number" doesn't match the newest version.
Implements #305

### DIFF
--- a/app/routes.py
+++ b/app/routes.py
@@ -21,10 +21,15 @@ from app.utils.session import generate_user_key, valid_user_session
 from bs4 import BeautifulSoup as bsoup
 from flask import jsonify, make_response, request, redirect, render_template, \
     send_file, session, url_for
-from requests import exceptions
+from requests import exceptions, get
 
 # Load DDG bang json files only on init
 bang_json = json.load(open(app.config['BANG_FILE']))
+
+# Get the newest version of WHOOGLE
+link_releases = "https://github.com/benbusby/whoogle-search/releases"
+update = bsoup(get(link_releases).text, "html.parser")
+newest_version = update.select_one('[class="Link--primary"]').string[1:]
 
 
 def auth_required(f):
@@ -137,6 +142,7 @@ def index():
         return render_template('error.html', error_message=error_message)
 
     return render_template('index.html',
+                           newest_version=newest_version,
                            languages=app.config['LANGUAGES'],
                            countries=app.config['COUNTRIES'],
                            themes=app.config['THEMES'],
@@ -261,6 +267,7 @@ def search():
 
     return render_template(
         'display.html',
+        newest_version=newest_version,
         query=urlparse.unquote(query),
         search_type=search_util.search_type,
         config=g.user_config,
--- a/app/routes.py
+++ b/app/routes.py
@@ -30,10 +30,11 @@ bang_json = json.load(open(app.config['BANG_FILE']))
 # Check the newest version of WHOOGLE
 update = bsoup(get(app.config['RELEASES_URL']).text, 'html.parser')
 newest_version = update.select_one('[class="Link--primary"]').string[1:]
-current_version = int(''.join(filter(str.isdigit, app.config['VERSION_NUMBER'])))
+current_version = int(''.join(filter(str.isdigit,
+                                     app.config['VERSION_NUMBER'])))
 newest_version = int(''.join(filter(str.isdigit, newest_version)))
-newest_version = newest_version if (current_version
-                                    < newest_version) else ''
+newest_version = '' if current_version >= newest_version \
+                    else newest_version
 
 
 def auth_required(f):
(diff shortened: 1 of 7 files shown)


app/routes.py
```diff
@@ -21,11 +21,16 @@
...
 from flask import jsonify, make_response, request, redirect, render_template, \
     send_file, session, url_for
-from requests import exceptions
+from requests import exceptions, get
 
 # Load DDG bang json files only on init
 bang_json = json.load(open(app.config['BANG_FILE']))
 
+# Get the newest version of WHOOGLE
+link_releases = "https://github.com/benbusby/whoogle-search/releases"
```
benbusby, line 30: App related constants like this should be in `app/__init__.py`, for example: `app.config['RELEASES_URL']` or something along those lines.

app/routes.py
```diff
@@ -21,11 +21,16 @@
...
+update = bsoup(get(link_releases).text, "html.parser")
```
benbusby, line 31: Small change here -- I try to uniformly use single quote strings throughout the repo

```suggestion
update = bsoup(get(link_releases).text, 'html.parser')
```

This will also have to be updated to reflect the change noted above as well

app/templates/footer.html
```diff
@@ -0,0 +1,10 @@
+<link href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.0.0-beta2/css/all.min.css" rel="stylesheet">
```
benbusby, line 1: I'd prefer not to import third party libs like font-awesome. So far the project is getting by fine with just vanilla JS and internal styling only. 

Could we just use a unicode character like "⬆️" or "🟢" or something? It isn't as pretty, but would work well enough imo.

app/templates/footer.html
```diff
@@ -0,0 +1,10 @@
...
+<footer>
+    <p class="footer">
+        Whoogle Search v{{ version_number }} ||
+        <a id="gh-link" href="https://github.com/benbusby/whoogle-search">{{ translation['github-link'] }}</a>
+        {% if newest_version != version_number %}
```
benbusby, line 6: I think this could be improved a bit to allow users using branch builds with potentially a higher version number (i.e. branch build is `v0.6.2`, latest release is `v0.6.1`) to not trigger this check.

Off the top of my head, you could convert the version to an int easily enough with something like:

```py
int(''.join(filter(str.isdigit, version_str)))
```

and then check if the latest version is actually higher than the current version, but you should probably perform that logic outside of the template and then pass in the result as a template var.

benbusby (changes requested):
Thanks @vacom13, great work. A few small adjustments to make, then this should be good to merge.

vacom13:
Okay. I will get to it.

benbusby (approved):
Awesome, looks good. Thanks again!

vacom13:
@benbusby I want to level up helping 😅. For now, been doing uk beginner level stuff. Well I am a newbie but I want to learn so how else can I contribute or are there resources u could refer to me?

benbusby:
@vacom13 the work you've been doing has been great, and super helpful, even if it might feel like beginner level stuff. If you want to contribute more, you can always look through the available issues and just start picking any that sound interesting to you. There's a few items that come to mind that might be a bit more difficult, but would definitely be well received by users I think, such as:

- Allow defining a list of proxy connections instead of just one, and cycling through them if any are blocked by Captcha
    - See #383 (would need to factor in both latency and availability -- low latency but blocked by Captcha proxies would need to be skipped)
    - Would also need to ensure that this doesn't hinder app performance too much.
- Adding "nice to have" features to the results view. These all require some sort of integration with an external service, so you'd need to be careful with how you're implementing it. Things like ratelimiting would need to be investigated prior to implementation. 
    - Show hours for a business if that's the first result (see #301)
    - Implement a more full featured currency conversion (see #129) -- this one is closed, but I'd be willing to re-open if there's a straightforward way to implement without requiring a third party JS library or something like that. Ideally it would be a new endpoint within Whoogle for converting currency, with a regex to detect if the search is related to currency. I'm thinking `/convert?from=usd&to=gbp&amount=1.99`, with a really simple UI embedded as the first result, similar to what you did for the "My IP" card
 
I can try thinking of other things, but it depends on what your interests are. Like I said, I'm happy to assign any of the available issues to you. You're doing great with contributing so far, it's been nice having someone else around to help clear out the backlog of issues 😄

benbusby:
You could also look into deploying to other services that users have requested in the past, seen here: #62

Those might feel a bit too simple, but provide a lot of value for people who have preferred deployment environments that aren't being addressed. One of the core benefits of Whoogle (imo) is the ease of deployment, so the more support for other deployment options, the better.

vacom13:
Okay. Great. I will just look into the issues then and try to figure stuff out on the go I guess.

vacom13:
@benbusby I was thinking of maybe giving the implementation of the full currency conversion a try. Something like what google does right? I can use like my own javascript tho right?

Issue #520, opened 2021-11-02
[BUG] Internal server error for some image searches when View Image option is enabled


fabianschilling:
**Describe the bug**
Certain image searches result in an Internal Server Error when the environment variable `WHOOGLE_CONFIG_VIEW_IMAGE=1` is set. The bug occurs due to a footer that is added to the bottom of the image search results on the Google website:

> Some results may have been removed under data protection law in Europe. [Learn more](https://www.google.com/policies/faq?hl=en)

**To Reproduce**
1. Input something in the search field that is likely to be restricted under European data protection law, e.g. "joe biden"
2. Click Search
3. Click on the "IMAGES" tab
4. See Internal Server Error

**Deployment Method**
- [ ] Heroku (one-click deploy)
- [ ] Docker
- [x] `run` executable
- [ ] pip/pipx
- [ ] Other: [describe setup]

**Version of Whoogle Search**
- [ ] Latest build from [source] (i.e. GitHub, Docker Hub, pip, etc)
- [x] Version [v0.6.0]
- [ ] Not sure

**Desktop (please complete the following information):**
 - OS: [macOS Monterey 12.0.1]
 - Browser [Firefox]
 - Version [93.0]

PR #521, opened 2021-11-02
Fixes internal server error

fabianschilling:
Fixes https://github.com/benbusby/whoogle-search/issues/520.


benbusby (approved):
Thank you!

Issue #322, opened 2021-05-22
[FEATURE] country/region selection in result instead of configuration


eddydc:
**Describe the feature you'd like to see added**
Currently a country can only be selected via the configuration. 

I would prefer that one would select the country/region when the results are shown instead of changing the config each time. A good reference for this is startpage.com

**Additional context**

bruvv:
I too think we could improve a lot with dual languages. For example. I am from Germany, speaking English and living in The Netherlands. I use a german VPN and my language search option is set to english. 
When I am searching for `volkswagen` I want to get the dutch site for the car company them selfs: `https://www.volkswagen.nl/` but this is what I am getting:
![afbeelding](https://user-images.githubusercontent.com/3063928/121145519-92b02400-c82e-11eb-8eb3-58ee40288aad.png)

Google:
![afbeelding](https://user-images.githubusercontent.com/3063928/121145564-9e034f80-c82e-11eb-98c0-bead11003a6a.png)

How can we fix this for people that do not only want english results but do speak english :P ?

benbusby:
@bruvv 

> How can we fix this for people that do not only want english results but do speak english :P ?

Can you share the URL for the Google search you performed? The URL parameters would tell me more about how the results were fetched than a screenshot. 

That being said, a simple change of the "Search Language" from English to Dutch returns the same results for Whoogle as the results from your Google search, at least in my quick test. In your Whoogle screenshot you have your Search Language set to English, which is why only English language websites are returned (i.e. `volkswagen.com` instead of `volkswagen.nl`).

eddydc:
I never encountered such an option in Google itself and one needs to choose explicitly the country via the URL eg Google.de or Google.be. Otherwise Google chooses the site based on ip location.
Therefore I was referring to startpage.com which let's you choose the country after the initial search if you don't want the ip location based results. I guess they change the Google URL in the background.

bruvv:
> 
> Can you share the URL for the Google search you performed? The URL parameters would tell me more about how the results were fetched than a screenshot.
```
[...]

Issue #192, opened 2021-02-09
[FEATURE] add google search tools function


somebody-somewhere-over-the-rainbow:
**Describe the feature you'd like to see added**
Add search tools capabilities currently available in google search (limit time / age of search results and various settings of image search results)

benbusby:
Time limiting is available with the `:past <year|month|week|day>` field at the end of a search query (i.e. `"darknet diaries episodes :past week"`). Image search result tools can still use the same query based filtering that regular Google search allows, just without the UI (so things like `filetype:png` should still work in an image search).

Does this satisfy what you're looking for? Or is this more a request for implementing the dropdown filter UI that Google implements for search?

alexw1982:
I am talking about all the options available in the highlighted section:
![grafik](https://user-images.githubusercontent.com/11628284/108707632-9c787a80-7510-11eb-8d19-77c9e471cd85.png)

If I am able to fully get there (e.g. having only images with a transparent background) via query filters in the search bar - that would satisfy my personal need. I was unable to find documentation on all the available filters (link?)

Long term I would still think that the dropdown UI would me easier to use for most people ...

geckolinux:
@benbusby Ahh, thanks, I wasn't aware of the `:past` filter. But I agree that a dropdown selector would be more discoverable, in fact, I would even recommend that it be always visible, not hidden under a menu as Google does.
By the way, congrats on making Whoogle, I really appreciate it.

Don-Swanson:
Given the previous response from Ben, is there still a request here, or can this be closed?

Issue #136, opened 2020-10-22
[BUG] Shopping tab styling/filtering issues


saltven:
**Describe the bug**
When shopping tab is clicked, two headers (logo and search bar) are displayed instead of the usual header. Plus, the whole page for shopping is in a expanded mobile view but not in a desktop view. 

![Screenshot_2020-10-22 test - Whoogle Search](https://user-images.githubusercontent.com/39688612/96878560-95335880-14ad-11eb-9108-fb0b7d2e4dc9.png)

benbusby:
While looking into this issue, I realized the shopping tab actually may need it's own separate filtering step, as there's a few things wrong with it.

Beyond just the styling issues -- the double header, missing `max width` and `margin: 0 auto` properties, etc -- a majority of the results link to Google adclick redirects (i.e. `google.com/aclk?...`) and don't contain any immediately discernible parameter for determining the true URL of the intended page. I pushed a hotfix to remove all links to the shopping tab temporarily until this can be addressed more thoroughly.

timknowlden:
Hi, is there any timeframe to bringing back the shopping tab, its the only reason I have to use googles website sometimes.

Don-Swanson:
Given how complex the Google Shopping search has become and how variable it is from search to search, there doesn't appear to be a feasible way to bring this feature back.

PR #535, opened 2021-11-17
Improve rendering of the tabs 

### LINKED ISSUE #298: [FEATURE] Include category tabs (all/images/news/etc) in header
**Describe the bug**
The image tab has a different layout (CSS?) from the other tabs like "all", "videos", "news". The text of the menu is bigger on the image tab.
Moreover, in the image tab I'm not able to access maps nor books, those links aren't displayed.

image
image

Is this intended? It's a bit disturbing to have a sudden change of layout like this.

**To Reproduce**
1. Do a search.
2. Click on the "image" link in the menu.
3. See different layout from the previous page

**Deployment Method**
- [ ] Docker
- [ ] `run` executable
- [x] pip/pipx
- [ ] Other: [describe setup]

**Version of Whoogle Search**
- [x] Latest build from [source] (i.e. GitHub, Docker Hub, pip, etc)
- [ ] Not sure

 - OS: ArchLinux
 - Browser: Firefox
 - Version: 88.0 (64-bit)

**Additional context**

jacr13:
The rendering is similar in all the tabs. 

In addition, the tab disposition is the same across all the tabs.

If the translation is not present, the default names will be displayed (in English), some effort needs to be made here to add translation for all supported languages. It shouldn't be very hard to translate since they are very basic words.

This probably fixes #298. I intentionally removed the shopping tab because it requires some effort to render correctly (#136).

The header should be extended to add an extra row with different parameters to choice from (#322 and possibly #192)


app/utils/search.py
```diff
@@ -122,6 +122,7 @@ def generate_response(self) -> str:
                                self.request_params,
                                self.config,
                                content_filter.near)
+        self.full_query = full_query
```
benbusby, line 125: This doesn't appear to get used anywhere

```suggestion
```

app/utils/results.py
```diff
@@ -222,3 +223,41 @@ def add_ip_card(html_soup: BeautifulSoup, ip: str) -> BeautifulSoup:
...
+    tabs = copy.deepcopy(tabs)
+    for tab_id, tab_content in tabs.items():
+        # update name to desired language
+        if tab_id in translation:
+            tab_content['name'] = translation[tab_id]
+
+        # update href with query
+        query = full_query
+        if '&tbm=' in full_query:
+            query = full_query.replace(f"&tbm={search_type}", '')
```
benbusby, line 252: Can be simplified to

```suggestion
        query = full_query.replace(f'&tbm={search_type}', '')
```

app/templates/header.html
```diff
@@ -67,6 +67,25 @@
             </form>
         </div>
     </header>
+    <div>
+      <div class="desktop-tap-div">
```
benbusby, line 71: I assume this is a typo?

```suggestion
      <div class="desktop-tab-div">
```

app/static/css/header.css
```diff
@@ -106,3 +106,95 @@ header {
     width: 100%;
     flex: 1
 }
+
+.desktop-tap-div {
```
benbusby, line 110: Typo here (mentioned below as well)

```suggestion
.desktop-tab-div {
```

app/filter.py
```diff
@@ -277,6 +288,28 @@ def update_styling(self, soup) -> None:
         except AttributeError:
             pass
 
+        # Fix body max with on images tab
```
benbusby, line 291: Can you elaborate what's going on here? I'm not sure I follow.

app/filter.py
```diff
@@ -277,6 +288,28 @@ def update_styling(self, soup) -> None:
...
+            css_html_tag = (
+                'html{'
+                'font-family: Roboto, Helvetica Neue, Arial, sans-serif;'
+                'font-size: 14px;'
+                'line-height: 20px;'
+                'text-size-adjust: 100%;'
+                'color: #3c4043;'
+                'word-wrap: break-word;'
+                'background-color: #fff;'
+                '}'
```
benbusby, line 305: Why is this necessary? It doesn't seem to affect the page from what I can tell, and things like `background-color` and `color` are being ignored altogether. I tried removing it and couldn't tell any difference.

app/templates/header.html
```diff
@@ -67,6 +67,25 @@
...
+                              <a class="eZt8xd" href="{{ tab_content['href'] }}">{{ tab_content['name'] }}</a>
+                          {% endif %}
+                      {% endfor %}
+                      <div class="FElbsf"></div>
+                  </div>
+              </div>
+          </div>
+      </div>
+      <div class="" id="s">
+  </div>
```
benbusby, line 88: Is there a `mobile-tab-div` in the works?

app/static/settings/translations.json
```diff
@@ -34,7 +34,13 @@
...
         "light": "light",
         "dark": "dark",
-        "system": "system"
+        "system": "system",
+        "all": "All",
+        "images": "Images",
+        "maps": "Maps",
+        "videos": "Videos",
+        "news": "News",
+        "books": "Books"
```
benbusby, line 43: Equal translations are enforced in the tests, so these have to be translated to the existing languages. I'd like to automate this in the future, but haven't looked into it yet. 

That being said, I can take care of these. Whenever the feature is ready to merge, I'll go through and add the remaining translations before merging.

app/static/css/header.css
```diff
@@ -106,3 +106,95 @@ header {
...
+    border-radius: 0 0 8px 8px;
+    box-shadow: 0 2px 3px rgba(32, 33, 36, 0.18);
+    margin-bottom: 10px;
+    overflow: hidden;
+}
+
+.bRsWnc {
+    background-color: #fff;
```
benbusby, line 118: Is this being applied? It's likely getting overwritten by the Whoogle theme. 

```suggestion
    background-color: #fff;
```

app/static/css/header.css
```diff
@@ -106,3 +106,95 @@ header {
...
+    font-size: 16px;
+    padding: 0 0 0 8px;
+    flex: 1;
+    height: 35px;
+    outline: none;
+    border: none;
+    width: 100%;
+    -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
+    overflow: hidden;
+}
```
benbusby, line 200: From the class names, I'm guessing that these were extracted from the styles that Google includes with the tab bar. 

Would it be too much trouble to update the names of these classes to be more accurate? I'd like to avoid scenarios where we're using obfuscated class names provided by Google, since that ends up making the project harder to maintain if we need to tweak any of these later.

benbusby (changes requested):
Awesome @jacr13, thanks for working on this feature. On desktop it seems to work really well. I did a quick review, the main things are:

- Are you working on a mobile tab replacement as well? Currently the mobile tab bar is removed but not replaced by anything
- Some styling elements are potentially not necessary, or need to be updated.

jacr13:
Thanks @benbusby for reviewing the code!

This is a first version and of course it needs some rework. 
For the mobile version I have not done any testing, I will do when addressing the other issues!

I will not have too much time the next two weeks, so I may take a couple of weeks to fix yours remarks and implement the rest.

app/utils/search.py
jacr13: It is used in the routes.py file when I call get_tab_contents [here](https://github.com/jacr13/whoogle-search/blob/7a86519b84bf3e232ae1268893e87e962e2b987a/app/routes.py#L263).

app/utils/results.py
jacr13: Good point!

app/filter.py
jacr13: The max-width returned to the images tab is 700 (for body tag) which is different on all other tabs (736) so we need to replace it. The html styling is also missing and is needed to have the exact same formatting.

app/filter.py
jacr13: The colors and background colors can be removed, the others are needed to have a uniform styling across the tabs (It is mainly needed to fix the text formatting). It is only used in the images tab when the view image option is disabled!

app/templates/header.html
jacr13: I will work on this !

app/static/settings/translations.json
jacr13: Yes, I saw that the test didn't pass due to the keys not being the same on all languages, but I have just filled the English to have the "reference" keys.

app/static/css/header.css
jacr13: Maybe not, I have just taken all the css related to the tabs, I need to do some cleaning.

app/static/css/header.css
jacr13: Exact!

You are right, will fix it.

jacr13:
I have added support for the mobile version and fixed the issues you highlighted before.

The translation still needs to be done for all the languages (except English).

benbusby:
> I have added support for the mobile version and fixed the issues you highlighted before.
> 
> The translation still needs to be done for all the languages (except English).

Awesome! Will review this week.

benbusby:
Just FYI -- I'm holding off on reviewing this for a bit longer since Google is in the middle of changing how results are formatted, and I want to make sure their formatting changes don't break this feature. I'll be sure to review once it seems like the new result formatting changes they're making are finished/permanent.

DUOLabs333:
Problem I've been having -- between switching tabs, for a brief second, it switches to some font (maybe Times New Roman), before going back to the default font.

PR #536, opened 2021-11-18
Currency Conversion

vacom13:
Added a fully fledged currency conversion feature. 
Implements #129


app/static/js/currency.js
```diff
@@ -0,0 +1,11 @@
...
+    // id's for currency input boxes
+    id1 = "cb" + n1; 
+    id2 = "cb" + n2;
+
+    // getting the value of the input box that just got filled
+    var input_box = document.getElementById(id1).value;
+
+    // updating the other input box after conversion
+    document.getElementById(id2).value = ((input_box * conversion_factor).toFixed(2));
+}
```
benbusby, line 11: ES6/formatting updates:

```suggestion
const convert = (n1, n2, conversionFactor) => {
    // id's for currency input boxes
    let id1 = "cb" + n1; 
    let id2 = "cb" + n2;

    // getting the value of the input box that just got filled
    let inputBox = document.getElementById(id1).value;

    // updating the other input box after conversion
    document.getElementById(id2).value = ((inputBox * conversionFactor).toFixed(2));
}
```

app/utils/results.py
```diff
@@ -223,3 +223,90 @@ def add_ip_card(html_soup: BeautifulSoup, ip: str) -> BeautifulSoup:
...
+
+    """
+    soup = BeautifulSoup(response, 'html.parser')
+    a = soup.find(class_='ZINbbc xpd O9g5cc uUPGi')
+    tag = a.find('a')
+    if tag and tag['href'] == 'https://g.co/gfd':
+        currency1 = soup.find(class_="xUrNXd UMOHqf")
+        currency2 = soup.select_one('.kCrYT'
+                                    ' .BNeawe.iBp4i.AP7Wnd '
+                                    '.BNeawe.iBp4i.AP7Wnd')
```
benbusby, line 245: Relying on this many obfuscated classes from Google makes me a bit nervous, since they could change any of these and kill this feature as a result...but I'm not sure of a better alternative. 

Could you maybe try finding the `a` tag with the link related to currency, using something like:

```py
currency_link = soup.find('a', {'href': 'https://g.co/gfd'})
```

and then maybe if that's found, traverse up the `div` tree using `<element>.parent` (swapping `<element>` for the previous parent result) until one has the `ZINbbc` class? `ZINbbc` seems to be the main class that Google uses for result divs, so if/when they change that one, it would be an easy thing to replace with whatever they change it to.

app/utils/results.py
```diff
@@ -223,3 +223,90 @@ def add_ip_card(html_soup: BeautifulSoup, ip: str) -> BeautifulSoup:
...
+        currency1 = currency1.text.rstrip('=').split(' ', 1)
+        currency2 = currency2.text.split(' ', 1)
+        return [float(currency1[0]), currency1[1],
+                float(currency2[0]), currency2[1]]
+    return None
+
+
+def add_currency_b(soup: BeautifulSoup,
```
benbusby, line 253: Could probably use a better name...maybe `add_currency_card`?

app/routes.py
```diff
@@ -15,7 +15,8 @@
 from app.request import Request, TorError
 from app.utils.bangs import resolve_bang
 from app.utils.misc import read_config_bool, get_client_ip
-from app.utils.results import add_ip_card
+from app.utils.results import add_ip_card, check_currency
+from app.utils.results import add_currency_b
 from app.utils.results import bold_search_terms
```
benbusby, line 20: These could be condensed into a single import from `app.utils.results`. You'll probably have to break it up into multiple lines though (with `\`)

app/utils/results.py
```diff
@@ -223,3 +223,90 @@ def add_ip_card(html_soup: BeautifulSoup, ip: str) -> BeautifulSoup:
...
+
+
+def check_currency(response: str):
+    """Check whether the results have currency conversion
+
+    Args:
+        response: Search query Result
+
+    Returns:
+        list
```
benbusby, line 235: Since the returned list contains a specific set of values that you're expecting in `add_currency_details`, it might be good to document those here. Or it might be better to a key-value dict so that you can reference them by more fitting names (see below)

app/utils/results.py
```diff
@@ -223,3 +223,90 @@ def add_ip_card(html_soup: BeautifulSoup, ip: str) -> BeautifulSoup:
...
+    tag = a.find('a')
+    if tag and tag['href'] == 'https://g.co/gfd':
+        currency1 = soup.find(class_="xUrNXd UMOHqf")
+        currency2 = soup.select_one('.kCrYT'
+                                    ' .BNeawe.iBp4i.AP7Wnd '
+                                    '.BNeawe.iBp4i.AP7Wnd')
+        currency1 = currency1.text.rstrip('=').split(' ', 1)
+        currency2 = currency2.text.split(' ', 1)
+        return [float(currency1[0]), currency1[1],
+                float(currency2[0]), currency2[1]]
```
benbusby, line 249: Could be better as a dict, so that these at least can be referenced by name:

```suggestion
        return {
            'currencyValue1': float(currency1[0]),
            'currencyLabel1': currency1[1],
            ...
        }
```

or something like that. I generally try to make variable names as descriptive as possible in cases like these.

app/utils/results.py
```diff
@@ -223,3 +223,90 @@ def add_ip_card(html_soup: BeautifulSoup, ip: str) -> BeautifulSoup:
...
+
+    Returns:
+        BeautifulSoup
+    """
+    # Element before which the code will be changed
+    # (This is the 'disclaimer' link)
+    element1 = soup.select_one('[class="nXE3Ob"]')
+
+    # Creating the conversion factor
+    conversion_factor = conversion_details[0]/conversion_details[2]
```
benbusby, line 271: Remember to update these to reference dict values if you change the return type of the previous method.

app/utils/results.py
```diff
@@ -223,3 +223,90 @@ def add_ip_card(html_soup: BeautifulSoup, ip: str) -> BeautifulSoup:
...
+
+    # Creating a new div for the input boxes
+    conversion_box = soup.new_tag('div')
+    conversion_box['class'] = 'conversion_box'
+
+    # Currency to be converted from
+    input_box1 = soup.new_tag('input')
+    input_box1['id'] = 'cb1'
+    input_box1['type'] = 'number'
+    input_box1['class'] = 'cb Hg3mWc'
```
benbusby, line 281: Do we have to call this class `Hg3mWc`? Or can we just extract the styles from that class, rename it something clearer (`currency_element` or something, idk) and include the styling in the main css?

app/utils/results.py
```diff
@@ -223,3 +223,90 @@ def add_ip_card(html_soup: BeautifulSoup, ip: str) -> BeautifulSoup:
...
+        soup: Parsed search result
+        conversion_details: list of currency
+        related information
+
+    Returns:
+        BeautifulSoup
+    """
+    # Element before which the code will be changed
+    # (This is the 'disclaimer' link)
+    element1 = soup.select_one('[class="nXE3Ob"]')
```
benbusby, line 268: I think this is fine since it's a single class, but it kinda makes me wonder if we should start maintaining an enum of classes used by Google, and then reference them using clearer names. Then it would also only require modifying one file if the class names ever change. So something like:

```py
class GClass(Enum):
    result_class = 'ZINbbc'
    currency_disclaimer = 'nXE30b'
    # etc
```

and then reference in these instances using something like:

```py
element1 = soup.select_one(f'[class="{GClass.currency_disclaimer}"]')
```

Don't worry about doing any of that in this PR, I can handle it later. I'm more just curious if you had any thoughts?

benbusby (changes requested):
Thanks @vacom13! I did a quick review, but haven't had a chance to manually test out how it works yet. 

My main concern is the reliance on the obfuscated class names from Google. These are liable to change at any point in the future, so I try to not rely on them unless necessary. In this case, I think at least one solution would be finding the currency disclaimer link and then traversing up the tree until an expected class is found. That way there's only one class to change later (if it does change) rather than a large number of classes.

Otherwise it looks good so far!

vacom13:
@benbusby sure I will look at it

app/utils/results.py
vacom13: Hey @benbusby. Sorry, I have been down with the flu and so didn't get time to check these out before 😅. I do think it would be a good idea to maintain an enum of classes used by google.

app/utils/results.py
vacom13: I will rename it. I just had it to keep it consistent with the search results but keeping it clearer would be better yes.

app/utils/results.py
vacom13: Thanks!

app/static/js/currency.js
vacom13: Um I am not so well versed with js.  Could you explain as to why we would take `const convert = ` and then `=>`?

app/utils/results.py
vacom13: Yes. I will try to do so. Don't know why I took the harder route 😅. Well maybe I could ensure that the text present is actually just disclaimer?  Instead of checking for parent?

app/utils/results.py
vacom13: Yes. Makes it consistent.

app/routes.py
vacom13: oh okay. I seem to be having a lot of issues with pep8. Do you have any plugins or add-ons for code editors which I could use to make it easier?

app/utils/results.py
vacom13: Yes. I will switch to a dictionary.

app/utils/results.py
vacom13: Maybe after merging this pr, I could take up that issue. Since I am the one who has been using these classes a bunch!

app/static/js/currency.js
benbusby: The "fat arrow" syntax is just part of ES6 -- [see here](https://www.javascripttutorial.net/es6/javascript-arrow-function/).

Declaring as `const` makes the function immutable, which helps avoid any issues (albeit unlikely) where other parts of the js code change this function.

app/routes.py
benbusby: Which editor do you use? I use Vim, so I have [this plugin](https://github.com/tell-k/vim-autopep8) installed and configured to format on save. I also use Pycharm occasionally though, and that editor automatically shows pep8 formatting errors (although I don't think it enforces that you fix them). 

I'm sure there's other plugins if you don't use either of those editors, but I can't recommend any in particular. Alternatively you could set up a [pre-commit hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) to run the same pep8 check that the repo uses -- `pycodestyle app/*` (although that approach is a lot more work). 

Ultimately I wouldn't worry about it if you're getting errors from the pep8 part of the build. As long as it gets fixed before merge it's no big deal.

app/utils/results.py
benbusby: No worries! Hope you're feeling better!

app/routes.py
vacom13: I have been using Pycharm but it doesn't seem to care about the number of characters in the line. Like it allows me to type more than 72 characters. Maybe I need to set the limit myself? I also use sublime text editor. 

Well I have used sublime for the most part. Pycharm creates its .idea files and a virtual env so I find that annoying, deleting those before merging 😅.

 But then I am sure there is a way to avoid adding those files but I just don't have that much git knowledge I guess

vacom13:
@benbusby I have done all the changes. But it shows that the branch has conflicts. How do I got about this?

vacom13:
I will just open a new PR I suppose?

vacom13:
Okay, I think I got it 😅

vacom13:
I went ahead and did away with most of those google classes.  They are just present in the CSS for now, I think. If you have an idea about how to go about those, I will gladly make the changes.

app/utils/results.py
```diff
@@ -223,3 +223,101 @@ def add_ip_card(html_soup: BeautifulSoup, ip: str) -> BeautifulSoup:
...
+    soup = BeautifulSoup(response, 'html.parser')
+    currency_link = soup.find('a', {'href': 'https://g.co/gfd'})
+    if currency_link and currency_link.text.lower() == 'disclaimer':
+        for i in range(4):
+            currency_link = currency_link.parent
+        currency1 = currency_link
+        for i in range(5):
+            currency1 = currency1.parent
+        for i in range(2):
+            currency1 = currency1.previous_sibling
```
benbusby, line 247: I think looking for a specific class name would actually be a better approach -- if we switch to traversing up N number of parents, and the layout for the conversion card returned by Google gets changed at some point, it'll be more effort to determine how deep the child element is nested. As long as we keep it to one class name, it should be easy enough to maintain in the future. 

Something like (pseudocode, not tested):

```py
while currency_link.parent and 'ZINbbc' not in currency_link.attrs['class']:
    currency_link = currency_link.parent
```

benbusby (changes requested):
Hey @vacom13, this looks good! One recommendation though (see below), and I'm also getting a 500 error if I search for an exact conversion (i.e. "11 eur to usd").

vacom13:
@benbusby okay will tackle this right away 😅

app/utils/results.py
```diff
@@ -223,3 +223,101 @@ def add_ip_card(html_soup: BeautifulSoup, ip: str) -> BeautifulSoup:
...
+    Args:
+        response: Search query Result
+
+    Returns:
+        dict: Consists of currency names and values
+
+    """
+    soup = BeautifulSoup(response, 'html.parser')
+    currency_link = soup.find('a', {'href': 'https://g.co/gfd'})
+    if currency_link and currency_link.text.lower() == 'disclaimer':
```
benbusby, line 240: Also, does this still work when the search language is changed? If not, it's probably good enough to just check if the currency link exists.

vacom13: I will check it out

vacom13:
@benbusby So I was checking the languages and well, I am running into a value error. In the example below as you can see that the floating numbers are using commas.

image

image

benbusby:
@vacom13 ah. You could probably just perform a quick replace on the value before converting to float, a la `.replace(',', '.')` (sorry if that doesn't format properly, I'm not at my computer).

vacom13:
@benbusby The format was fine. No worries. I did run into more issues using the `.replace()`. If I asked for the output of lets say `10,000 usd to inr`, the output (in English) came out to be `752,415.00 Indian Rupee`. The `,` in the result would then get replaced by a `.`.  Similarly, I checked for the language Portuguese and they used `,` in place of `.` to indicate the start of the decimal numbers and `.` instead of `,` (Basically the roles of `,` and `.` were reversed).  However, I think I have managed to make it work now.

app/utils/results.py
```diff
@@ -223,3 +223,107 @@ def add_ip_card(html_soup: BeautifulSoup, ip: str) -> BeautifulSoup:
...
+    conversion_box = soup.new_tag('div')
+    conversion_box['class'] = 'conversion_box'
+
+    # Currency to be converted from
+    input_box1 = soup.new_tag('input')
+    input_box1['id'] = 'cb1'
+    input_box1['type'] = 'number'
+    input_box1['class'] = 'cb'
+    input_box1['value'] = conversion_details['currencyValue1']
+    input_box1['oninput'] = f'Convert(1, 2, {1 / conversion_factor})'
```
benbusby, line 300: Needs to be updated to match new function name

```suggestion
    input_box1['oninput'] = f'convert(1, 2, {1 / conversion_factor})'
```

app/utils/results.py
```diff
@@ -223,3 +223,107 @@ def add_ip_card(html_soup: BeautifulSoup, ip: str) -> BeautifulSoup:
...
+
+    br = soup.new_tag('br')
+
+    # Currency to be converted to
+    input_box2 = soup.new_tag('input')
+    input_box2['id'] = 'cb2'
+    input_box2['type'] = 'number'
+    input_box2['class'] = 'cb'
+    input_box2['value'] = conversion_details['currencyValue2']
+    input_box2['oninput'] = f'Convert(2, 1, {conversion_factor})'
```
benbusby, line 315: Same as above

```suggestion
    input_box2['oninput'] = f'convert(2, 1, {conversion_factor})'
```

benbusby (changes requested):
OK @vacom13, two more small changes (really just one change repeated twice) and then I'll merge. Thanks again for taking care of this! It seems to work really well, nice job.

app/utils/results.py
vacom13: 😅

vacom13:
@benbusby I well fetched and merged the changes before committing my own and this seems to be giving a test failed error 😢. Not sure if it has anything to do with my code.

benbusby:
@vacom13 yeah, that test fails randomly. I've been meaning to re-write it so that it's more reliable, but haven't gotten around to it yet. I just re-ran the test suite though and everything is passing now.

benbusby (approved):

vacom13:
Okay Great!

FrankelJb:
Sorry for the necrobump...

Experiencing this on 0.7.3. Running the Docker image on k3s. The issue seems to be resolved, started occurring again today. Logs below:

```
File "/usr/local/lib/python3.11/site-packages/flask/app.py", line 2446, in wsgi_app
response = self.full_dispatch_request()
^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/flask/app.py", line 1951, in full_dispatch_request
rv = self.handle_user_exception(e)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/flask/app.py", line 1820, in handle_user_exception
reraise(exc_type, exc_value, tb)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/flask/_compat.py", line 39, in reraise
raise value
^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/flask/app.py", line 1949, in full_dispatch_request
rv = self.dispatch_request()
^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/flask/app.py", line 1935, in dispatch_request
return self.view_functions[rule.endpoint](**req.view_args)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/whoogle/app/routes.py", line 94, in decorated
return f(*args, **kwargs)
^^^^^^^^^^^^^^^^^^
File "/whoogle/app/routes.py", line 53, in decorated
return f(*args, **kwargs)
^^^^^^^^^^^^^^^^^^
File "/whoogle/app/routes.py", line 361, in search
conversion = check_currency(str(response))
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/whoogle/app/utils/results.py", line 268, in check_currency
while 'class' not in currency_link.attrs or \
^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'attrs'
```

PR #573, opened 2021-12-11
alpine docker image

roylez:
I have to remove libcurl4-openssl-dev from list of dependencies as there is no such package in alpine, but the app seems to run just fine without it.

Uncompressed alpine image size is one third of current slim build.

```
> docker images |grep whoogle
whoogle      alpine       64514ef8ff7d   9 minutes ago    84.2MB
whoogle      ubuntu       d16e591bb7c6   20 minutes ago   245MB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,64 +1,57 @@
-FROM python:3.8-slim as builder
+FROM python:3.8-alpine as builder
 
-RUN apt-get update && apt-get install -y \
-    build-essential \
+RUN apk --update add \
+    build-base \
     libxml2-dev \
     libxslt-dev \
-    libssl-dev \
+    openssl-dev \
     libffi-dev
-
+ 
 COPY requirements.txt .
 
 RUN pip install --prefix /install --no-warn-script-location --no-cache-dir -r requirements.txt
 
-FROM python:3.8-slim
+FROM python:3.8-alpine
 
-RUN apt-get update && apt-get install -y \
-    libcurl4-openssl-dev \
-    tor \
-    curl \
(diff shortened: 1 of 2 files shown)


martadinata666:
And debian based build wise, add `--no-install-recommends` will stripe out un-needed packages significantly. After building, the package only need `libcurl`. 👍🏼

roylez:
An alpine based docker image is available at `roylez/whoogle-search` for whoever wants to try for now.

Honestly, it could be further reduced by replacing curl with libcurl, removing bash dependency by changing shebang from `#!/bin/bash` to `#!/bin/sh`, but switching to alpine provides the most of gain here anyway.

benbusby (approved):
Looks good, thanks!

gdm85:
I got this failure which I believe is due to Alpine:
```
--> 6e90f98ecd8
[1/2] STEP 3/5: COPY requirements.txt .
--> 21bdf0938fe
[1/2] STEP 4/5: RUN pip install --upgrade pip
Requirement already satisfied: pip in /usr/local/lib/python3.11/site-packages (21.2.4)
Collecting pip
  Downloading pip-22.0.4-py3-none-any.whl (2.1 MB)
Installing collected packages: pip
  Attempting uninstall: pip
    Found existing installation: pip 21.2.4
    Uninstalling pip-21.2.4:
ERROR: Could not install packages due to an OSError: [Errno 39] Directory not empty: 'webencodings'
[2/2] STEP 1/40: FROM python:3.11.0a5-alpine
```

Generally I experience a lot of issues with Alpine-based images so I switched back to using the previous approach with Debian.

Issue #610, opened 2022-01-12
[FEATURE] No autocomplete request when autocomplete disabled


nakoo:
**Describe the feature you'd like to see added**
A short description of the feature, and what it would accomplish.

`whoogle.env` supports changing visibility of autocomplete. But this feature makes sense by making no `/autocomplete` request. This could help reducing the server load.

**Additional context**

![2022-01-12_18-07](https://user-images.githubusercontent.com/4975021/149197211-39ba1d19-fb32-42bb-b352-5bc121b0bd25.png)

Albonycal:
so `/autocomplete` should be disabled when the ENV variable is set to 0 ?

nakoo:
> so `/autocomplete` should be disabled when the ENV variable is set to 0 ?

The current implementation is no. It still requests `/autocomplete` even though the results don't show up.

DUOLabs333:
I can implement this -- it's not that hard. I just need to know @benbusby -- how are preferences exposed in the program (ex. `WHOOGLE_AUTOCOMPLETE`)?

DUOLabs333:
#611 should fix this.

PR #611, opened 2022-01-14
Disable autocomplete.js if `WHOOGLE_AUTOCOMPLETE=0`

DUOLabs333:


DUOLabs333:
Should fix #610

benbusby (changes requested):
The treatment of this environment variable needs to be inverted. `WHOOGLE_AUTOCOMPLETE` should only ever used to disable the feature. I'd rather not have an entire feature suddenly disappear for the majority of users who are accustomed to it being enabled by default.

DUOLabs333:
But it is -- if it doesn't exist, it is 1 by default.

benbusby:
Ah I see. The title of the PR threw me off and I hadn't looked at the code much yet. Will review now.

app/routes.py
```diff
@@ -37,6 +37,8 @@
 newest_version = '' if current_version >= newest_version \
     else newest_version
 
+ac_var='WHOOGLE_AUTOCOMPLETE'
+autocomplete_enabled=os.getenv(ac_var,'1')
```
benbusby, line 41: pep8 fixes:

```suggestion
ac_var = 'WHOOGLE_AUTOCOMPLETE'
autocomplete_enabled = os.getenv(ac_var, '1')
```

app/templates/display.html
```diff
@@ -35,7 +35,9 @@
 {{ response|safe }}
 </body>
 {% include 'footer.html' %}
-<script src="{{ cb_url('autocomplete.js') }}"></script>
+{% if autocomplete_enabled == '1' %}
+        <script src="{{ cb_url('autocomplete.js') }}"></script>
+    {% endif %}
```
benbusby, line 40: Indentation issue
```suggestion
{% if autocomplete_enabled == '1' %}
    <script src="{{ cb_url('autocomplete.js') }}"></script>
{% endif %}
```

benbusby (changes requested):
Looks good, just a couple minor formatting fixes before I merge.

app/routes.py
```diff
@@ -37,6 +37,8 @@
 newest_version = '' if current_version >= newest_version \
     else newest_version
 
+ac_var = 'WHOOGLE_AUTOCOMPLETE'
+autocomplete_enabled = os.getenv(ac_var, '1')
```
benbusby, line 41: pep8 also enforces 2 line breaks before function defs

```suggestion
autocomplete_enabled = os.getenv(ac_var, '1')

```

DUOLabs333:
So, is it ready to be merged?

benbusby (approved):

PR #612, opened 2022-01-15
Improve formatting of collapsible infobox

nityy:
Some long infoboxes (typically movies, artists) are treated as collapsibles due to having more than 5 child divs. This PR improves the subtitle (typically year, length, genre, profession, links etc.) display by putting it in parentheses and appending to title. Also sets its font weight to normal.

A instance is running [here](https://whoogle-test.fly.dev/) with the changes until this is merged.

before
**Before**
***
after
**After**


benbusby (approved):
Ah, this has been bugging me for a while, just hadn't gotten around to fixing it. Thank you so much!

PR #640, opened 2022-02-04
Remove blocked sites from "Images" search bar

### LINKED ISSUE #638: [BUG] If specified blocked sites, it shows up if you switch to "Images"
**Describe the bug**
If you have sites blocked and search, if you switch to "Images", in the search bar, there are "-site:...". This does not happens if you search on Images directly, and does not happen on any other tab.

**To Reproduce**
1. Specify blocked sites in `whoogle.env`
2. Search something. 
3. Switch to "Images"

**Deployment Method**
- [ ] Docker
- [x] `run` executable
- [ ] pip/pipx
- [ ] Other: [describe setup]

**Version of Whoogle Search**
- [x] Latest build from [source] (i.e. GitHub, Docker Hub, pip, etc)
- [ ] Not sure

 - OS: [e.g. iOS] Arch Linux
 - Browser [e.g. chrome, safari] Firefox
 - Version [e.g. 22] 88

**Smartphone (please complete the following information):**

**Additional context**

DUOLabs333:
Fixes #638

### DIFF
--- a/app/filter.py
+++ b/app/filter.py
@@ -363,6 +363,9 @@ class Filter:
 
         # get some tags that are unchanged between mobile and pc versions
         search_input = soup.find_all('td', attrs={'class': "O4cRJf"})[0]
+        
+        search_input=BeautifulSoup(re.sub(r'(\s)-site:[a-zA-Z0-9_\.]+', '', 
+        str(search_input)),'lxml')
         search_options = soup.find_all('div', attrs={'class': "M7pB2"})[0]
         cor_suggested = soup.find_all('table', attrs={'class': "By0U9"})
         next_pages = soup.find_all('table', attrs={'class': "uZgmoc"})[0]


app/filter.py
```diff
@@ -363,6 +363,9 @@ def view_image(self, soup) -> BeautifulSoup:
 
         # get some tags that are unchanged between mobile and pc versions
         search_input = soup.find_all('td', attrs={'class': "O4cRJf"})[0]
+
+        search_input = BeautifulSoup(re.sub(r'(\s)-site:[a-zA-Z0-9_\.]+', '',
+                                     str(search_input)), 'lxml')
```
benbusby, line 368: Should use a different parser here -- iirc `lxml` isn't included by default.

```suggestion
                                     str(search_input)), 'html.parser')
```

benbusby (changes requested):
Looks good, but it only works if the user also has the `View Full Size Images` option enabled. It might be a little hard to remove it for the image results page without that config enabled, since we're not using the standard header template for image results.

app/filter.py
DUOLabs333: Weird, it works for me in my virtualenv

benbusby:
I might hold off on merging this until I finish reviewing/merging #535, since that should give us more general control over what appears in the search bar + header.

DUOLabs333:
If all tabs use the same header, etc. through #535, then this might not be necessary.

benbusby:
Confirmed fixed by #535, closing this one since that PR just got merged.

PR #650, opened 2022-02-09
Add gowogle.voring.me as public instance

ThatOneCalculator:
Many of the listed instances are down and outdated. gOwOgle (which I run) is alive and up-to-date.


Albonycal:
hmm, my instance isn't dead?  
I'm using it RN
it was down for a few hours tho

PR #651, opened 2022-02-10
Fix error with `remove_site_blocks` in the Images tab

DUOLabs333:


app/filter.py
```diff
@@ -125,6 +125,9 @@ def clean(self, soup) -> BeautifulSoup:
     def remove_site_blocks(self, soup) -> None:
         if not self.config.block:
             return
+
+        if not soup.body:
+            return
```
benbusby, line 130:
```suggestion
        if not self.config.block or not soup.body:
            return
```

benbusby (approved):

PR #659, opened 2022-02-15
Give Accept-Language div its own class

nityy:
Minor change. Noticed this when I wanted to hide accept-language but not get-only for my personal instance.


benbusby (approved):
Oops, good catch. Thanks!

PR #667, opened 2022-02-23
Configure setup() using setup.cfg

CyberTailor:
Setuptools documentation:
https://setuptools.pypa.io/en/latest/userguide/declarative_config.html


benbusby (approved):
Thanks!

PR #682, opened 2022-03-14
Anonymous view

benbusby:
The current `/window` endpoint is used as a proxy currently, but only
for removing Javascript from the result page. This expands the existing
functionality to allow users to proxy search result pages through their
Whoogle instance.

Note that this feature will likely be merged in without being 100%
finished, since I'm unsure of how users intend to use the feature. With
some community testing, I'll get a better idea of A) how many people are
actually going to use this feature, and B) what aspects of the feature
are useful or not useful.

### DIFF
--- a/app/filter.py
+++ b/app/filter.py
@@ -264,7 +264,7 @@ class Filter:
                 # enabled
                 parent.decompose()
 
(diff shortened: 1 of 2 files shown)


Albonycal:
:eyes: This would be awesome!

PR #715, opened 2022-04-06
Add support for relative search results

### LINKED ISSUE #441: [FEATURE] "base url" support

Support to use a base url, for example: `https://this-is-an-example-domain.com/baseurl/`, `https://this-is-an-shared-domain.com/whoogle/`

**Additional context**

Currently, we can't use a shared domain to only proxy requests with some prefix to a whoogle instance, the requests will be redirected to use the root path (`https://this-is-an-example-domain.com/`).

gdm85:
If you are running whoogle on a non-root URI (e.g. `/something/whoogle`) then results will not work.

This PR make some changes to the souped'up results so that URLs will be correctly relativised.

Fixes #441

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,7 @@ RUN mkdir -p $config_dir
 RUN chmod a+w $config_dir
 VOLUME $config_dir
 
+ARG url_prefix=''
 ARG username=''
 ARG password=''
 ARG proxyuser=''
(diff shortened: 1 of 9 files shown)


gdm85:
This PR works well on top of 0a96134eab9794912548484731e67b0a80f2f01a but I realised that on `main` there are many changes (related to session, mostly) that make this not working anymore.

benbusby (approved):
I made one small addition to support relative paths for internal Flask redirects using Waitress' `url_prefix` arg. The value can be set using the `WHOOGLE_URL_PREFIX` env var, which will allow internal redirects (such as the session validation you mentioned) to also follow the correct path.

Issue #723, opened 2022-04-15
[BUG] anon-view KeyError


glitsj16:
**Describe the bug**
Introduction of [Support proxying results through Whoogle (aka "anonymous view")](https://github.com/benbusby/whoogle-search/commit/9317d9217f6d2ac39938ee402b9330f54f2f87dd) throws a KeyError in results.py:

```
File "/home/glitsj16/whoogle-search/app/utils/results.py", line 212, in append_anon_view
av_link.string = f'{translation["anon-view"]}'
KeyError: 'anon-view'
```

**To Reproduce**
1. Build from git main
2. Set 'WHOOGLE_CONFIG_ANON_VIEW=1'
3. Run a search

**Deployment Method**
- [ ] Heroku (one-click deploy)
- [ ] Docker
- [ x] `run` executable
- [ ] pip/pipx
- [ ] Other: [describe setup]

**Version of Whoogle Search**
- [ x] Latest build from [source] (i.e. GitHub, Docker Hub, pip, etc)
- [ ] Version [version number]
- [ ] Not sure

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Browser: Firefox
 - Version: 99.0.1

PR #724, opened 2022-04-15
Fix 'anon-view' KeyError

glitsj16:
Fix for https://github.com/benbusby/whoogle-search/issues/723.

### DIFF
--- a/app/utils/results.py
+++ b/app/utils/results.py
@@ -209,7 +209,7 @@ def append_anon_view(result: BeautifulSoup, config: Config) -> None:
     translation = current_app.config['TRANSLATIONS'][
        config.get_localization_lang()
     ]
-    av_link.string = f'{translation["anon-view"]}'
+    av_link.string = f'{translation["config-anon-view"]}'
     av_link['class'] = 'anon-view'
     result.append(av_link)
 


benbusby (approved):

benbusby:
Hey @glitsj16, I just reverted this change in 7688c1a23326da1ebb37283ba0b58b05043731e5 because the "anon-view" key is actually the correct value to use in search results. "config-anon-view" should only be used on the home page config menu. Can you re-try with the latest code and see if you're still getting the error? If so, could you also share which language you have configured for your instance?

glitsj16:
Hi @benbusby. I've done another test. My configured language in whoogle.env is `
WHOOGLE_CONFIG_LANGUAGE=lang_en` (as it has always been). Without referencing `WHOOGLE_CONFIG_ANON_VIEW` in my whoogle.env my instance is running without errors. As soon as I set `WHOOGLE_CONFIG_ANON_VIEW=1` the formerly reported error is thrown after running a search. The web interface shows
```
Internal Server Error

The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.
```
and on the server I can see the exact same error message as before...

It still does so after upgrading my instance to the latest git code. Perhaps I made a mistake in setting the relevant env var, not sure. I used GH search and can only see that env var in [config.py](https://github.com/benbusby/whoogle-search/blob/9317d9217f6d2ac39938ee402b9330f54f2f87dd/app/models/config.py#L31). Hope this clears up what I reported before. Feel free to ask if there's anything else I can do to help.

benbusby:
Hmmm, that's very strange. Could you do a quick sanity check to see if `app/static/settings/translations.json` contains the "anon-view" key on your end? I can't figure out why that one key doesn't appear for you, especially since apparently "config-anon-view" works for you, which was part of the same commit to that file. I'm wondering if there's an issue with pulling the correct translation mapping, but if your language is set to `lang_en` then it shouldn't be a problem...

I'll keep looking into it as well. Could you also see if enabling anonymous view from the home page config also causes the problem if `WHOOGLE_CONFIG_ANON_VIEW` is not set in whoogle.env?

glitsj16:
Very strange indeed. During the weekend I have some more time to investigate what could be causing this on my end. For now I can confirm the sanity check: `"config-anon-view": "Show Anonymous View Links",` is contained in app/static/settings/translations.json. When enabling anonymous view from the home page (without setting WHOOGLE_CONFIG_ANON_VIEW) I get the same behavior: internal server error... I really appreciate your efforts to get to the bottom of it and will report back here tomorrow or sunday after another full reinstall of whoogle-search.

glitsj16:
@benbusby Aha, `problem solved`! Your last comment on `translations.json` did it. At first I didn't fully grasp that you were asking if I had _BOTH_ "anon-view" and "config-anon-view" in that file. I must have missed an update or messed-up otherwise, but after a full reinstall of whoogle-search from latest git it is now working as intended on all my instances. Thanks again for looking into this and my apologies for what turned out to be 'just user error noise'.

PR #746, opened 2022-05-04
Tor auth

MadcowOG:
Added the ability for user to authenticate tor control ports with passwords. It could be a better alternative to cookies, as passwords doesn't require changing file permissions.

### DIFF
--- a/app/request.py
+++ b/app/request.py
@@ -8,6 +8,7 @@ import urllib.parse as urlparse
 import os
 from stem import Signal, SocketError
 from stem.control import Controller
+from stem.connection import authenticate_cookie, authenticate_password
 
 MAPS_URL = 'https://maps.google.com/maps'
 AUTOCOMPLETE_URL = ('https://suggestqueries.google.com/'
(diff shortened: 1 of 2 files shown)


MadcowOG:
I apologize for taking up so much workflow runs, I'm still trying to figure things out.

nakoo:
For privacy perspective, I don't think it's a good idea to implement password instead of cookie. In general, I don't understand why we need to give user permission to change the tor port by risking server security.

MadcowOG:
I understand. I think this could provide more options to the user, as neither option is directly more secure than the other. With the way I'm thinking this should be set up, it would restrict whoogle's access to the file system much more. Which from a security standpoint could be more pleasant.

MadcowOG:
> the tor control port by risking server security.

What about this method could be a security risk?

nakoo:
> > the tor control port by risking server security.
> 
> What about this method could be a security risk?

I read your code. The only thing this code do is deciding whether server needs to enable Tor or not. I do believe this is something for server administrator, not user. It's unnecessary extra step to disable Tor connection. Manging password makes hard to manage Docker container secure.

MadcowOG:
I am confused

> The only thing this code do is deciding whether server needs to enable Tor or not. It's unnecessary extra step to disable Tor connection.
Didn't the code previous to my changes do this already? I just made it so the code determines this factor through passwords.

> Manging password makes hard to secure Docker container too.
My code doesn't change anything with the Docker container, since by default cookies are only enabled in torrc in Docker, it doesn't even use password authentication in Docker, correct. I think you are right though passwords shouldn't be used in Docker.

MadcowOG:
>The only thing this code do is deciding whether server needs to enable
Tor or not. It's unnecessary extra step to disable Tor connection.

Sorry, do you mean the attempt to always authenticate passwords with multiple try and except blocks, is unnecessary? If so I agree, I will get on that.

MadcowOG:
>The only thing this code do is deciding whether server needs to enable Tor or not. It's unnecessary extra step to disable Tor connection.

@nakoo I think I understand what you were talking about here. Did my recent commits fix the issue with unnecessary extra steps?

benbusby (approved):
Thanks!

PR #748, opened 2022-05-05
Fix link in window

invis-z:
In anonymous view the link was replaced to /window?location=&{url} which is incompatable with WHOOGLE_URL_PREFIX


benbusby (approved):
Good catch, thanks!

PR #751, opened 2022-05-09
change Hebrew from lang_iw to lang_he

ShlomiD83:
set Hebrew to lang_he instead of lang_iw which makes no sense.


benbusby (commented):
These aren't decided by Whoogle, so they can't be modified. The values in this file encapsulate what Google will accept for the language parameter in the search url. 

I agree that your change makes more sense visually, but this would actually break Hebrew language support unfortunately.

ShlomiD83:
> These aren't decided by Whoogle, so they can't be modified. The values in this file encapsulate what Google will accept for the language parameter in the search url.
> 
> I agree that your change makes more sense visually, but this would actually break Hebrew language support unfortunately.

Realy? I've changed it on my Whoogle instance and it didn't break anything,

Issue #760, opened 2022-05-17
[BUG] Images do not load


SangeloDev:
**Describe the bug**
Images don't load when browsing. 
When opening Debug console, they point to ``localhost:port/element`` when behind NginX proxy.

**To Reproduce**
1. Use NginX as proxy and route domain.tld to localhost:port
2. Search anything
3. Click Images
4. See error (Images don't load)

**Deployment Method**
- [ ] Heroku (one-click deploy)
- [X] Docker
- [ ] `run` executable
- [ ] pip/pipx
- [ ] Other: [describe setup]

**Version of Whoogle Search**
- [X] Latest build from Docker: 0.7.2
- [ ] Version [version number]
- [ ] Not sure

**Desktop (please complete the following information):**
 - OS: Arch
 - Browser: Firefox Development Edition
 - Version 101.0b5

**Smartphone (please complete the following information):**
 - Device: OnePlus 8 pro
 - OS: EvolutionX ``Snow-v2-6.3.1-instantnoodlep / Android 12``
 - Browser Firefox Nightly
 - Version 102.0a1

**Additional context**
If there is any, I'll add it after an edit.

DUOLabs333:
It seems that the image is not being routed through whoogle. This may be a result of f5d599e7.

PR #734, opened 2022-04-26
Remove "/" before endpoints & tags

invis-z:
Fix #733

PR #762, opened 2022-05-19
Fix image view

invis-z:
I previously removed the / without noticing it was part of a string replacement in #734 This caused the href of "View Image" contain a leading "/" which is wrong. This PR added it back.

### DIFF
--- a/app/filter.py
+++ b/app/filter.py
@@ -533,7 +533,7 @@ class Filter:
                 continue
 
             img_url = urlparse.unquote(urls[0].replace(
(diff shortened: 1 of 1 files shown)


DUOLabs333:
This should fix #760.

benbusby (approved):

DUOLabs333:
Huh, in that case, I can't reproduce that issue.

SangeloDev:
#774 also seems to be broken, if you want, I could show my configuration to help with reproducing the issue?

PR #764, opened 2022-05-21
added instance whoogle.privacydev.net

PrivacyDevel:


benbusby (commented):
Looks good, but the instance also needs to be added to misc/instances.txt

benbusby (approved):

PR #769, opened 2022-05-25
Fix urls not open in new tab

### LINKED ISSUE #761: "Open Links in New Tab" setting no longer working
It appears the "Open Links in New Tab" setting has stopped working recently. I use the Docker image at benbusby/whoogle-search:latest . In the benbusby/whoogle-search:0.7.2 image (built 21 days ago) this setting functions normally.

jacr13:
Simple fix to allow again urls to open in new tabs.

Fixes #761

### DIFF
--- a/app/filter.py
+++ b/app/filter.py
@@ -435,6 +435,8 @@ class Filter:
             link['href'] = 'https://google.com' + q
         elif link['href'].startswith('/url'):
             link['href'] = q
+            if self.config.new_tab:
+                link['target'] = '_blank'
         elif q.startswith('https://accounts.google.com'):
             # Remove Sign-in link
             link.decompose()
--- a/app/filter.py
+++ b/app/filter.py
@@ -433,10 +433,6 @@ class Filter:
             # Internal google links (i.e. mail, maps, etc) should still
             # be forwarded to Google
             link['href'] = 'https://google.com' + q
-        elif link['href'].startswith('/url'):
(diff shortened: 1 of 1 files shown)


DUOLabs333:
You should really be fixing https://github.com/benbusby/whoogle-search/commit/0f6226ce51f08747eef1c45e0acfe5cb94e69b69

benbusby (approved):
So I looked into this a bit, and the cause of the problem is actually that this section of handling result links starting with `/url` shouldn't have been committed. These result links are already being filtered/handled later in the if-elif-else chain (which includes modifying the link to open in a new tab if configured that way), but since this block came before, it caused the proper filtering method to not get executed. So while your fix would've addressed the problem, I just pushed a change to just remove this section altogether, since it isn't actually needed.

Thanks!

benbusby:
Also @DUOLabs333 the `/window` endpoint is only used for anonymous view links, whereas this bug that's being reported affects even non-anonymous view links, so I think that's a separate issue. Although I'll add that I'm not sure why there needs to be a leading slash for anonymous view links -- it seems to work in my deployments, and I was under the impression that introducing a leading slash breaks instances that are being hosted behind a subpath of a domain (i.e. `mydomain.com/whoogle/` being the root level of the instance. In this case, if a leading slash were to be added for the `window` endpoint, it would redirect to `mydomain.com/window`, not `mydomain.com/whoogle/window`)

PR #785, opened 2022-06-14
Fix a crash with protected Tor control port

CyberTailor:


CyberTailor:
Kinda similar to #746 but simpler :)

MadcowOG:
This doesn't work, send_tor_signal is used for more than just to check for Tor availability. I made this same mistake in my pull request, too. We do need to authenticate with the Tor control port because of a new identity request.

This condition can be found at around line 300 in [request.py](./app/request.py):
``` Python
# Validate Tor conn and request new identity if the last one failed
        if self.tor and not send_tor_signal(
                Signal.NEWNYM if attempt > 0 else Signal.HEARTBEAT):
            raise TorError(
                "Tor was previously enabled, but the connection has been "
                "dropped. Please check your Tor configuration and try again.",
                disable=True)
```
We use this to ask Tor for a new circuit in the case google starts rejecting our requests or hit us with a captcha.

CyberTailor:
Oops, missed it. Now `AuthentificationError` sets `TOR_AVAILABLE` to 0.

benbusby (approved):

PR #801, opened 2022-06-29
Tor password newline fix

MadcowOG:
Sometimes a newline character is picked up when parsing control.conf. So just strip it when we encounter it.


app/request.py
```diff
@@ -57,6 +57,8 @@ def send_tor_signal(signal: Signal) -> bool:
                     for line in conf:
                         pass
                     secret = line
+                    if '\n' in line:
+                        secret = line.strip('\n')
```
benbusby, line 61: Could be simplified to:
```suggestion
                    secret = line.strip('\n')
```

PR #807, opened 2022-07-01
added instance (search.wef.lol)

J2D9:


benbusby (changes requested):
The instance should also be added to misc/instances.txt so that Farside can find the new instance as well.

J2D9:
#808

benbusby (approved):

i-cant-git:
@benbusby This website now redirects to `affordableluxurybags . nu`

Albonycal:
this instance is now redirecting to some weird website `https://affordableluxurybags.nu/`
we should probably remove it from the list

Albonycal:
https://github.com/benbusby/whoogle-search/pull/860

PR #836, opened 2022-08-30
Update languages.json

CYAXXX:


benbusby:
Combined and merged with #837

Issue #817, opened 2022-07-17
[FEATURE] <Url to save settings instead of a cookie>


MariojGG:
In other privacy-oriented search engines there´s also the option of saving your preferences with a url instead of storing a cookie in your browser, so users that delete cookies each browsing session can use custom settings too.

Screenshots

SearX     -     https://imgbox.com/goW3iWq0

Qwant    -    https://imgbox.com/p2WlZwRA

Startpage  - https://imgbox.com/TNOHXExp

PR #842, opened 2022-09-07
Add preferences parameter to url

jacr13:
The preferences' parameter contain an encrypted version of the desired config of the user.

The encryption step is a choice and not required, as it is here a key is mandatory to show the new url in configuration.
If we decide to go without encryption, the preferences' parameter could work on different instances of whoogle. Maybe it is useful for people using farside to have it not encrypted?

The next possible step, would be to add a config variable to deactivate completely the cookies ?!

Fixes: #817 

Translation needs to be done.


benbusby (changes requested):
Thanks for working on this, this is great!

I do think that the encryption step should be optional, since that way we could pass preferences to other Whoogle instances when one is rate-limited. Whoogle config values aren't terribly sensitive anyways, so I don't think it would fit most users use cases to have these encrypted. 

We could also (as an additional feature for later, not in this PR) make it so that Whoogle preferences could be mapped to Searx preferences if all Whoogle instances get rate-limited.

I can go through and add translations as well once this is tweaked to allow unencrypted preferences.

jacr13:
I have made the changes to allow the user to set the encryption or not, the option will only work if a key has been previously set in the environment variables, as mentioned under the option.

As it is now, all instances can read unencrypted preferences and encrypted preferences can only be read by instances with the same key.

To know if a token is encrypted or unencrypted when sharing a token, I have set a flag on the token, if the token starts with :
 - 'e': the token is encrypted, and a key is needed to decode it, eg. 

> 127.0.0.1:5000/?preferences=**e**G2MDAGSqbfR8MSPS....

 - 'u': the token is unencrypted and can be shared and used with all the instances, eg. 

> 127.0.0.1:5000/?preferences=**u**G8EGACwKbCd6ME...

benbusby (approved):

PR #864, opened 2022-10-23
Updating Readme.md ; Heroku fix and Removed Some Non-Working Public Instances

hrishikeshp7:
Updated readme.md ; Heroku has stopped free plans so I have updated readme according to that . And some of the Public instances were not working  (Actually only one) . I have removed them from the list .


files18:
The removed public instance is https://whoogle.lunar.icu

Issue #845, opened 2022-09-13
[BUG] Search terms are not bold in Chinese results


haorenxu:
**Describe the bug**
Search terms are bold in English results, but not bold in Chinese results.

**To Reproduce**
Search anything in Chinese.

**Deployment Method**
- [ ] Heroku (one-click deploy)
- [x] Docker
- [ ] `run` executable
- [ ] pip/pipx
- [ ] Other: [describe setup]

**Version of Whoogle Search**
- [x] Latest build from [source] (i.e. GitHub, Docker Hub, pip, etc)
- [ ] Version [version number]
- [ ] Not sure

**Desktop (please complete the following information):**
 - OS: MacOS 12.5
 - Browser Safari
 - Version 15

**Smartphone (please complete the following information):**
 - Device: iPhone 13
 - OS: iOS 15
 - Browser Safari
 - Version 15

![E447AC38-BBCE-457E-A243-1C6ACAE02AA7_1_201_a](https://user-images.githubusercontent.com/33184148/189819928-40d9c7fd-8796-4462-9bd3-db4cbf6155d0.jpeg)
![33A7EFD0-C68A-4ECE-9643-0D8ADB43F9A3_1_201_a](https://user-images.githubusercontent.com/33184148/189819935-09a82c1c-1c35-4d80-888f-c98f1a7637c1.jpeg)

haorenxu:
@benbusby Thanks for you guys to fix this, but there are still some bugs.

Here is the result in searching "新聞" ("News" in English) on Google. We can see the keywords "新聞" are all red.
<img width="822" alt="Screen Shot 2022-11-14 at 5 06 57 PM" src="https://user-images.githubusercontent.com/33184148/201620552-468e588f-e8ac-44a1-a191-d1ed3d57e85a.png">

And here is the result on Whoogle, only one keyword (the first one) is bold, and the rest are not.
<img width="793" alt="Screen Shot 2022-11-14 at 5 07 04 PM" src="https://user-images.githubusercontent.com/33184148/201621546-d080aaee-bc08-4f18-ad6e-4ad9e2f33140.png">

PR #865, opened 2022-10-24
changes regex to replace special characters and not chinese characters. Issue-#845

Abir10101:
Updated regex matching where it is removing chinese letters as thinking special characters.


benbusby (approved):

haorenxu:
Hi, thanks for fixing this, but something is not working. Please see information provided in #845.

PR #876, opened 2022-11-02
Added Welsh Translation

curlpipe:
More info about the language:
https://en.wikipedia.org/wiki/Welsh_language


benbusby (approved):
Thank you!

PR #879, opened 2022-11-05
Include new ad filter keyword

xaabi6:
Hello, this new keyword will help recognizing more sponsored search results:

Sponsored result

This search was made from my private instance which is in Spain and uses Spanish language.

By the way, congratulations for this awesome project!


benbusby (approved):
Great! Thank you for your contribution!

Issue #908, opened 2022-12-15
[BUG] Results renders HTML elements


prashantrahul141:
when searching for any type of html element ( like input ) if the results contains something like `<input/>` it gets rendered on the results page.

1. search html elements
example: https://www.whoogle.click/search?q=image+type+input+html

**Version of Whoogle Search**
- the latest one whoogle.click is using.

**Desktop:**
 - OS: Windows
 - Browser Brave
 - Version 1.46.140

**Example:**
On Google:
![google](https://user-images.githubusercontent.com/59825803/207822561-405a1669-a248-4a85-bf4e-fd3d99396c51.png)

On Whoogle:
![whoogle](https://user-images.githubusercontent.com/59825803/207822632-2c68012a-ab19-4de1-b1aa-cc1e252d68c6.png)

![whoogle1](https://user-images.githubusercontent.com/59825803/207823283-e3a670fb-12a1-4b31-b7db-36df3beb56f4.png)

ahmad-alkadri:
Reproduced the bug from Whoogle on Docker by typing in `<Input>` on the search bar:

<img width="1465" alt="Screenshot 2022-12-22 at 11 54 32" src="https://user-images.githubusercontent.com/22837764/209119855-d0937049-5b86-4f50-8a74-e37d8951ff81.png">

> **Browser**: Firefox 107.0.1
> **OS**: MacOS 13.1
> **Machine**: Macbook Pro, M1 Pro chip

By inspecting the element, I can see that the problem is that the escape check/functionality is not applied there. I'm not sure without delving into the details, but I think, to fix this, simply adding an escape feature for the text passed into the HTML template would be sufficient.

Personally, it's a bit surprising to see this bug though; I've been using Whoogle for a while and never noticed this. 

---

PR #912, opened 2022-12-23
Fix: added a functionality to make sure escaped characters stay escaped.

ahmad-alkadri:
This PR is linked to the issue #908 which shows that, basically, Whoogle results render html characters unescaped. Here's a screenshot as referenced in the issue:

image

After checking, I found out that several points: 
+ the characters inside the `<div>` content tag from the search results (`getbody.text` in search.py) are already escaped, with `"<"` and `">"` characters converted into `"&lt;"` and `"&gt;"`, respectively
+ however, because the `getbody.text` then passed through several `bsoup` class, the escaped tag characters became unescaped.

To prevent this, I replaced `"&lt;"` and `"&gt;"` with `"andlt;"` and `"andgt;"`, respectively. This way, when the 'response' object get loaded to `bsoup` (which happens several times throughout the process between search.py and routes.py), `bsoup` will not unescape them. Finally, at the end, before the `responses` object sent to the `render_template` in `routes.py`, I simply replaced the `"andlt;"` and `"andgt;"` back to `"&lt;"` and `"&gt;"`.

Here's the screenshot from the search result on Whoogle following this fix:

screenshot-localhost_5000-2022 12 23-22_58_09

### DIFF
--- a/app/routes.py
+++ b/app/routes.py
@@ -284,7 +284,6 @@ def autocomplete():
         g.user_request.autocomplete(q) if not g.user_config.tor else []
     ])
 
-
 @app.route(f'/{Endpoint.search}', methods=['GET', 'POST'])
 @session_required
 @auth_required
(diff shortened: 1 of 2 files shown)


benbusby (approved):
Thanks!

ahmad-alkadri:
> Thanks!

You're welcome! It's a pleasure.

Issue #211, opened 2021-02-26
[FEATURE] anti-captcha support.


ghost:
with stuff that can get blocked easily, anti-captcha support would be huge. invidious has it implemented perfectly, and it allows public instances to be used with out any major rate-limiting. just an idea. thanks!

benbusby:
I do agree that would be a nice feature. I haven't encountered the captcha issue too much (thankfully) but I'd still like to eliminate it altogether.

My understanding of anti-captcha services at the moment is that the worthwhile ones aren't free, which unfortunately prevents it from being something I can universally implement in the code base. I suppose I could put the responsibility on the public instance maintainer to provide an API key for activating the service -- I'm not sure how Invidious and others handle it, but I assume it's something like that. 

In any case, I'll look into it. Thanks for the recommendation!

fres7h:
Invidious implements it very nicely,  I actually didn't know it was an option until talking with the developers. You'd just add a line in your config for your API key. I'm willing to pay if it means my instance can be used without worry of getting blocked, plus it's super cheap.

unixfox:
> I do agree that would be a nice feature. I haven't encountered the captcha issue too much (thankfully) but I'd still like to eliminate it altogether.
> 
> My understanding of anti-captcha services at the moment is that the worthwhile ones aren't free, which unfortunately prevents it from being something I can universally implement in the code base. I suppose I could put the responsibility on the public instance maintainer to provide an API key for activating the service -- I'm not sure how Invidious and others handle it, but I assume it's something like that.
> 
> In any case, I'll look into it. Thanks for the recommendation!

You can implement the anti-captcha API, it's not universal nor a standard, but it's very common and easy to clone.

A lot of projects provide an anti-captcha API clone like https://capmonster.cloud or mine which I plan to release it publicly as soon as I find it stable.

Albonycal:
Yea I'm also getting blocked by google...
This would be cool..
Any updates?
Thank you :D

maxdesalle:
Also having this issue on a DigitalOcean droplet.

randomwalk3141592:
Rather than an anti-captcha feature, can you make it so that Whoogle will actually display Google's captcha so that it can be solved?  Right now, the captcha is not displayed.

unixfox:
> Rather than an anti-captcha feature, can you make it so that Whoogle will actually display Google's captcha so that it can be solved?  Right now, the captcha is not displayed.

[...]

Issue #763, opened 2022-05-21
[BUG] <Instance gets ratelimited if more than 2-3 devices use it>


AzureB1te:
**Describe the bug**
When I use whoogle on mobile, there are no issues with it. I also use Kaffeine to keep it alive. But if I start using it on desktop, it gets ratelimites in a day or two

**To Reproduce**
1. Use whoogle on device #1
2. Use whoogle on device #2
3. Add your instance to Kaffeine (probably Heroku only)

**Deployment Method**
- [ ✓] Heroku (one-click deploy)
- [ ] Docker
- [ ] `run` executable
- [ ] pip/pipx
- [ ] Other: [describe setup]

**Version of Whoogle Search**
- [ ] Latest build from [source] (i.e. GitHub, Docker Hub, pip, etc)
- [ ] Version [version number]
- [ ✓] Not sure

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

bruvv:
Duplicate https://github.com/benbusby/whoogle-search/issues/707 have you searched before posting ?

Issue #707, opened 2022-04-01
[BUG] Instance has been ratelimited 


eddydc:
**Describe the bug**
Getting 

Error

Instance has been ratelimited

[Continue your search with Farside](https://farside.link/)
Whoogle:
https://farside.link/whoogle/search?q=dit is een test&lang_interface=lang_en&country=BE&theme=system

Searx:
https://farside.link/searx/search?q=dit is een test

[Return Home](https://synoogle:Hf4LhYGvsg@wilma.idizaai.be/home)

**To Reproduce**
Just do a search with Whoogle

**Deployment Method**
- [ ] Docker-compose

**Version of Whoogle Search**
lastest

**Desktop (please complete the following information):**
Win 10, latest FF

T3chHunt3r:
I too have this issue on version 0.7.0

ShlomiD83:
same issue for me.
can't resolved the issue no matter what I do.

sysoppl:
Same for me, just deployed and it's not working at all.

ironboy1:
Same on my side for version 0.7.1 and 0.7.2

bruvv:
This has been answered already: https://github.com/benbusby/whoogle-search/issues/211

pieslinger:
I am having rate limiting issues occasionally and I am the only user of whoogle running in a container. I do not see how #211 has answered this already. I would install this for more people in my life as well if this issue goes away. Is there just something I need to modify in the setup?

pieslinger:
I also only noticed it in 0.7.1 and 0.7.2 - although 0.7 was my first time using this. Just installed 0.7.3 tonight.

aw-jansen:
Same here. Getting the rate limiting message when connected via a NordVPN container. Would it still be a possibility to implement a workaround? On a different topic someone mentioned changing the query: https://github.com/searxng/searxng/issues/159

haorenxu:
Same here on 0.7.4 using a public IP with password. Getting limited for average one or two days. Restarting the docker can fix it, but the problem always appears again.

jojolll:
I use my own Whoogle instance (alone). Almost every day I reach the ratelimit after 10 to 15 searches during the day. 

I have the change with my LiveBox to be able with a simple click in the interface to renew my IP, that solves the problem for 1-2 days. 
It's still very annoying and makes the solution of a personal Whoogle server very unviable.

Albonycal:
[...]

PR #915, opened 2022-12-27
Attempt to bypass Google's CAPCHA using a third party service

Moist-Cat:
refs #211

Using a proxy or making any "suspicious" request triggers Google's reCAPTCHA v2 (with callback). I'm using https://deathbycaptcha.com/ to try to solve it.
I tested it using a blacklisted proxy, and the integration seems to be working fine.  The problem is, Google just sends another CAPTCHA after we solve the first one so I'm not 100% sure if I'm sending the token to the correct endpoint of if Google is rejecting my request for some reason.
Manually solving the CAPTCHA gives the same result.


benbusby (commented):
Thanks for your effort on this. I'm personally not the biggest fan of CAPTCHA solving services, so I can't test this myself. But as long as it doesn't interfere with core Whoogle functionality (which it doesn't appear to) then I'm open to merging it. Are you still working on it though? From your comment it sounds like this integration doesn't actually solve the CAPTCHA, but maybe I misinterpreted something.

Moist-Cat:
>are you still working on it though? From your comment it sounds like this integration doesn't actually solve the CAPTCHA, but maybe I misinterpreted something.

It does solve the CAPTCHA and sends it to Google's servers. Just that I'm not sure if it removes the rate limit because Google just sent another CAPTCHA when I solve the first one. This happened both testing it by hand on the site and internally on the application. That's why I'm not convinced it solves the issue.
Feel free to wait until someone with the rate-limiting issue tests the integration and confirms it's indeed working to merge the code because because I tested it with a free proxy and the results might not be the same.

>Looks like this doesn't get used anywhere. Did you mean to raise this exception instead of the AttributeError later on?
No, I meant to use this in the `solve` function but in the end I decided to just return False whenever we failed to solve the CAPTCHA.

dominickp:
I wonder if the captcha could be returned to the Whoogle user. If indeed solving a few of these causes Google's rate limiting to back off a bit that would be awesome.

Moist-Cat:
Every site protected by Google's reCAPTCHA has an unique site key that needs to be posted in the form. If the host doesn't match with the sitekey the CAPTCHA won't load.
We could:
- spoof the host by editing `/etc/hosts` to resolve `google.com` as `127.0.0.1`
- make a form to post the params to https://google.com/sorry/index and let the user solve the CAPTCHA

In theory this should work but i'm not sure if spoofing the host would be desirable. Maybe we will have to give the process sudo privileges to edit the hosts file before and after the CAPTCHA is solved.

4194304:
#707 and #763 are still working, which makes it so I have to switch Whoogle instances every few days which insanely annoying

Don-Swanson:
@Moist-Cat 
What is the current status on this?
I too am not a fan of CAPTCHA solving services. 
Is this still a WIP or is this working code? It's also 3 years old, and I know A lot changes in 3 years.

Moist-Cat:
The rate limit persists after solving the CAPTCHA. My guess is that there is a some detail missed when making the POST request with the solution.
Feel free to close this PR.

PR #918, opened 2022-12-30
adding line break between header options

fiestasiesta:
Reason for this is because on mobile view, "Time Period:" shows up on a separate line from its dropdown


benbusby (approved):
Thanks!

PR #925, opened 2023-01-04
Don't prepend services that have schemes with '//'

cazwacki:
This caused issues with hitting some of my local services; despite the URL being correctly configured, the prepended '//' would cause some URLs to work incorrectly. Tested with both https:// and http://-prefixed alt sites.


cazwacki:
Should resolve #926

benbusby (approved):

PR #927, opened 2023-01-05
Add two more ungovernable.men instances

dominickp:
Received a couple more IPv4 addresses so I'd like to add a couple more instances


benbusby (approved):
Great, thanks!

Issue #904, opened 2022-12-11
[BUG] Most of the search terms are not bold in Chinese results


whaler-ragweed:
**Describe the bug**
Most of the search terms are not bold in Chinese results.

Whoogle results:
<img width="819" alt="Screenshot 2022-12-11 at 6 43 14 PM" src="https://user-images.githubusercontent.com/33184148/206899969-b2d72332-7eee-44ca-88ec-b126492d361e.png">

Google results for reference:
<img width="936" alt="Screenshot 2022-12-11 at 6 43 21 PM" src="https://user-images.githubusercontent.com/33184148/206899990-c2edcc1f-399a-4624-8e94-a3fe589db406.png">

**To Reproduce**
Search "新聞" or any other term in Chinese.

**Deployment Method**
- [x] Docker

**Version of Whoogle Search**
- [x] Latest build from [source] (i.e. GitHub, Docker Hub, pip, etc)

**Desktop (please complete the following information):**
 - OS: MacOS and iOS
 - Browser: Safari

ahmad-alkadri:
I managed to duplicate this issue on my local and other public instances of Whoogle.

I think the problem stems from the regex expression on this line:

https://github.com/benbusby/whoogle-search/blob/ccf9f06f2f9b3064594006ac9b3dfc4762819912/app/utils/results.py#L74

Basically what this regex does is it looks for words matching the query, but only if the words stand on their own or there are no other characters before or after it (other than whitespaces). Then, if it found matches, the function will modify those matches with bold text.

This is why the characters `新聞` in the string `Google 新聞` will be made into bold like this: `Google <b>新聞</b>` while in `匯集了世界各地的新聞來源` it will not be touched at all.

I think a simple fix would be removing the `\b` in `\b((?![{{}}<>-]){target_word}(?![{{}}<>-]))\b`, so it'll become `((?![{{}}<>-]){target_word}(?![{{}}<>-]))`. I've tested this fix on my local instance and it seems to work for even the other Chinese characters:

![afterchange_regex_bold](https://user-images.githubusercontent.com/22837764/211117392-a7155658-c99c-4452-8613-e07c96cdd1f9.png)

Comparison with the search result from a public instance with the original regex expression:

![beforechange_regex_bold](https://user-images.githubusercontent.com/22837764/211117428-99d7f6c3-6308-4688-959a-57eca112fec9.png)

[...]

PR #928, opened 2023-01-08
Add a function to check if target_word contains CJK characters

ahmad-alkadri:
This PR is linked to Issue #904 which shows that Whoogle results do not render bold all of target words if they are Chinese characters.

Further investigations show similar behavior for Japanese (hiragana, katakana, kanji), and Korean (hangul syllables, hangul jamo) characters: not all of the words displayed on the result page are bolded.

To handle this, a function was added to check if `target_word` in `bold_search_terms.replace_any_case` contains Chinese, Korean, or Japanese characters and apply the regex that doesn't check for whitespace. This way, each search term would be bolded differently.

Screenshots of the search results after the commits linked to this PR:

screenshot-localhost_5000-2023 01 08-23_12_08

screenshot-localhost_5000-2023 01 08-23_14_26

screenshot-localhost_5000-2023 01 08-23_17_20

### DIFF
--- a/app/utils/results.py
+++ b/app/utils/results.py
@@ -44,6 +44,28 @@ SITE_ALTS = {
 }
 
 
+def contains_cjko(s: str) -> bool:
+    """This function check whether or not a string contains Chinese, Japanese,
+    or Korean characters. It employs regex and uses the u escape sequence to
+    match any character in the following Unicode ranges:
+    - 4e00-9fff: Chinese characters
+    - 3040-309f: Japanese hiragana
+    - 30a0-30ff: Japanese katakana
+    - 4e00-9faf: Japanese kanji
+    - ac00-d7af: Korean hangul syllables
+    - 1100-11ff: Korean hangul jamo
+
+    Args:
+        s (str): string to be checked
+
+    Returns:
+        bool: True if the input s contains the characters and False otherwise
+    """
+    return bool(
+        re.search(
+            r'[\u4e00-\u9fff\u3040-\u309f\u30a0-\u30ff\u4e00-\u9faf\uac00-\ud7af\u1100-\u11ff]', s))
+
+
 def bold_search_terms(response: str, query: str) -> BeautifulSoup:
     """Wraps all search terms in bold tags (<b>). If any terms are wrapped
     in quotes, only that exact phrase will be made bold.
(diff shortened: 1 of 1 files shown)


app/utils/results.py
```diff
@@ -44,6 +44,28 @@
...
+
+    Args:
+        s (str): string to be checked
+
+    Returns:
+        bool: True if the input s contains the characters and False otherwise
+    """
+    return bool(
+        re.search(
+            r'[\u4e00-\u9fff\u3040-\u309f\u30a0-\u30ff\u4e00-\u9faf\uac00-\ud7af\u1100-\u11ff]', s))
```
benbusby, line 66: I think I'd prefer if the unicode ranges were defined inline with the code instead of in the function documentation. That way it's a bit easier to visually parse if/when one of the ranges ever needs to be updated in the future, and the change would only have to be made to one spot. Thoughts?

```suggestion
def contains_cjko(s: str) -> bool:
    """This function check whether or not a string contains Chinese, Japanese,
    or Korean characters. It employs regex and uses the u escape sequence to
    match any character in a set of Unicode ranges.

    Args:
        s (str): string to be checked

    Returns:
        bool: True if the input s contains the characters and False otherwise
    """
    unicode_ranges = ('\u4e00-\u9fff' # Chinese characters
                      '\u3040-\u309f' # Japanese hiragana
                      '\u30a0-\u30ff' # Japanese katakana
                      '\u4e00-\u9faf' # Japanese kanji
                      '\uac00-\ud7af' # Korean hangul syllables
                      '\u1100-\u11ff' # Korean hangul jamo
                      )
    return bool(re.search(fr'[{unicode_ranges}]', s))
```

benbusby (commented):
Looks good to me, thanks for taking care of this! I had just one idea for a way to make the code a little easier to modify in the future (if needed), let me know your thoughts.

ahmad-alkadri:
> Looks good to me, thanks for taking care of this! I had just one idea for a way to make the code a little easier to modify in the future (if needed), let me know your thoughts.

Hi @benbusby; thanks for your input! I agree with your suggestion, and I've tested it also on several queries: 

screenshot-localhost_5000-2023 01 09-19_58_06

screenshot-localhost_5000-2023 01 09-19_59_20

It seems to work well. So I've committed it to the branch; would you like to test it also? I'm available for other follow-up discussions/modifications on this.

benbusby (approved):
Looks great, thank you again!

PR #940, opened 2023-01-26
Added Indonesian translation

ahmad-alkadri:
Edited the `translations.json` to add Indonesian translations as `lang_id`.

Apparently it's already initiated in [`app/static/settings/languages.json`](https://github.com/ahmad-alkadri/whoogle-search/blob/026714bb8420f2455e7741480997404c6e50f80c/app/static/settings/languages.json#L27) but its translations haven't been put in the `translations.json` before, thus I added it.

I've tested on my local instance and it seems to work well.

image

image

Open to discussions if you find any problem with it!


benbusby (approved):
Looks good, thanks for contributing!

ahmad-alkadri:
> Looks good, thanks for contributing!

You're welcome! Always happy to contribute here!

Issue #951, opened 2023-02-05
[Feature request] Cookie to save successful authorization


zoof:
Is it possible to have the Whoogle cookie remember successful authorization?  I enabled auth to prevent eventual discovery of outside users and it would be nice if the Whoogle cookie kept track of a successful authorization so that it's not necessary to sign in every browser startup.

zoof:
Works beautifully!  Thanks!

PR #964, opened 2023-02-22
Add authentication to cookie

jacr13:
Fixes: #951 

When authenticated, the cookie set will allow the user to stay connected even if the browser is restarted.


benbusby (approved):
Works great, thank you!

Issue #934, opened 2023-01-22
[FEATURE] Allow simple append to official style with WHOOGLE_CONFIG_STYLE instead of overwriting it


DUOLabs333:
**Describe the feature you'd like to see added**
Currently, setting `WHOOGLE_CONFIG_STYLE` overwrites the official style. However, in some cases, you may simply want to add something, rather than completely overwrite it. So, maybe have something like if `WHOOGLE_CONFIG_STYLE` doesn't start with `:root`, just append to the official style, else, overwrite it?

**Additional context**
This request comes from the fact that I want to change the width of the page (which should probably be its own setting) and the color of the logo without changing anything else on the page. The idea is that this: `WHOOGLE_CONFIG_STYLE="body {max-width: 100% !important; /* or whatever width you want */} #whoogle-w { fill: #685e79; } #whoogle-h { fill: #685e79; } #whoogle-o-1 { fill: #685e79; } #whoogle-o-2 { fill: #685e79; } #whoogle-g { fill: #685e79; } #whoogle-l { fill: #685e79; } #whoogle-e { fill: #685e79; } "` should work as expected.

A POC:
```
 self.style = open(os.path.join(app_config['STATIC_FOLDER'], 'css/variables.css')).read()
        if os.getenv('WHOOGLE_CONFIG_STYLE',"").startswith(":root"):
            self.style=os.environ["WHOOGLE_CONFIG_STYLE"]
        else:
            self.style+=(" "+os.getenv('WHOOGLE_CONFIG_STYLE',""))
```

PR #965, opened 2023-02-24
Add better handling of custom css

jacr13:
Fixes: #934

This pull request aims to solve exactly the problem discussed in the linked issue.

I decided to remove the "default" css from the configuration menu, as it is present in the referenced url. This is a choice, and it is straightforward to add it back. Please let me know your preference.

A possible drawback of the current approach is that it replaces the css rule by the new one, so if there are things that you want to keep from the old one, you need to specify them. This is important to keep in mind when changing variables inside :root for example.

### DIFF
--- a/app/models/config.py
+++ b/app/models/config.py
@@ -1,4 +1,5 @@
 from inspect import Attribute
+from typing import Optional
 from app.utils.misc import read_config_bool
 from flask import current_app
 import os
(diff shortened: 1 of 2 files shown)


app/models/config.py
```diff
@@ -88,6 +112,33 @@ def get_attrs(self):
...
+
+        modified_sheet = cssutils.parseString(self.style_modified)
+        for rule in modified_sheet:
+            rule_default = get_rule_for_selector(style_sheet,
+                                                 rule.selectorText)
+            # if modified rule is in default stylesheet, update it
+            if rule_default is not None:
+                # TODO: update this in a smarter way to handle :root better
+                # for now if we change a varialbe in :root all other default
+                # variables need to be also present
```
benbusby, line 135: The variables in `:root` can still be overridden with the html tag though, for example:

```css
html {
    --whoogle-page-bg: #f0f0f0 !important;
}
```

It might be good enough to just update the CSS wiki page to indicate that individual variables can be overridden that way.

benbusby (commented):
One small note, otherwise this is good to go. If you think updating the wiki is good enough to address the issue with `:root` vars, feel free to remove the TODO.

benbusby (approved):

PR #969, opened 2023-03-14
Fix a bug on remove_block_titles() and remove_block_url()

xatier:
Fix the exception `AttributeError: 'Filter' object has no attribute 'block_url'` introduced in this commit [1].

`self.block_title` and `self.block_url` were members of the Filter object[2], but not anymore after commit [1].

This bug can be reproduced with setting `WHOOGLE_CONFIG_BLOCK_URL` to a non-empty string.

[1] https://github.com/benbusby/whoogle-search/commit/10a15e06e1bf6f1a22d1625950d8c221d99d6a07
[2] https://github.com/benbusby/whoogle-search/commit/284a8102c85bdc69674952bc30a6cfacfb8b01ee

### DIFF
--- a/app/filter.py
+++ b/app/filter.py
@@ -225,7 +225,7 @@ class Filter:
     def remove_block_titles(self) -> None:
         if not self.main_divs or not self.config.block_title:
             return
-        block_title = re.compile(self.block_title)
+        block_title = re.compile(self.config.block_title)
         for div in [_ for _ in self.main_divs.find_all('div', recursive=True)]:
             block_divs = [_ for _ in div.find_all('h3', recursive=True)
                           if block_title.search(_.text) is not None]
@@ -234,7 +234,7 @@ class Filter:
     def remove_block_url(self) -> None:
         if not self.main_divs or not self.config.block_url:
             return
-        block_url = re.compile(self.block_url)
+        block_url = re.compile(self.config.block_url)
         for div in [_ for _ in self.main_divs.find_all('div', recursive=True)]:
             block_divs = [_ for _ in div.find_all('a', recursive=True)
                           if block_url.search(_.attrs['href']) is not None]


xatier:
@benbusby please kindly take a look.
I'm actually surprised that no previous bug reports for this one, perhaps we should add `WHOOGLE_CONFIG_BLOCK_URL` and `WHOOGLE_CONFIG_BLOCK_TITLE` to README?

benbusby (approved):
Thanks for fixing. I can't say I'm surprised that nobody has noticed until now though -- I was already under the impression that this feature wasn't used by very many people.

Regarding adding it to the README, I'm a bit unsure since it might be better to just remove the feature altogether in a future release. Both of these configuration values require the user to input a regex to make them work, which I don't think is very user friendly. Also there's already support for blocking multiple websites using a comma separated list, which is a lot easier than writing a regex, especially for non-technical users. 

Either way, if you want to add documentation for these to the README, feel free to open another PR. But I guess I'm leaning more towards just removing this functionality in the near future unless there's a compelling reason to keep it.

xatier:
Sounds good, I've opened https://github.com/benbusby/whoogle-search/pull/971 to update the README. Perhaps this feature was low usage due to the fact it was undocumented?

I'm aware the `WHOOGLE_CONFIG_BLOCK` setting, but I think the impl has some :bug:? `search_string` is not a regex? https://github.com/benbusby/whoogle-search/blob/main/app/filter.py#L205

```
>>> x = 'foo.com,bar.com'
>>> search_string = ' '.join(['-site:' + _ for _ in x.split(',')])
>>> search_string
'-site:foo.com -site:bar.com'
```

benbusby:
Not a bug really -- in this case the regex can just be the `-site:...` strings themselves so that BeautifulSoup will find partial matches. If we were to only pass the `-site:...` strings without `re.compile`, it wouldn't return any content from the result page since there wouldn't be any exact matches.

xatier:
I see, thanks for the clarification!

PR #1005, opened 2023-05-17
Update routes.py

Sasikumar00:


Sasikumar00:
Adding extra padding "==" to the encoding will eventually take the required padding and nullify if any extra padding is observed

Sasikumar00:
The above solution should prevent the invalid padding bug

PR #1012, opened 2023-05-31
Add color picker widget

ellieeet123:
Adds a color picker widget above search results when the keyword "color" or "colour" is in the query. Also slightly refactors the widget inserting code in the backend. 

Note - During the writing of this I was interrupted by some real-life drama and had to step away from it for a few months and kind of forgot what I was doing here, so the code's quality isn't amazing. If you want to (or want me to) feel free to improve it or ask me to. But as far as I know it works for now.

### DIFF
--- a/app/routes.py
+++ b/app/routes.py
@@ -355,8 +355,8 @@ def search():
         html_soup = bsoup(str(response), 'html.parser')
         if search_util.widget == 'ip':
             response = add_ip_card(html_soup, get_client_ip(request))
-        elif search_util.widget == 'calculator' and not 'nojs' in request.args:
-            response = add_calculator_card(html_soup)
+        elif not 'nojs' in request.args:
+            response = add_widget(html_soup, search_util.widget)
 
     # Update tabs content
     tabs = get_tabs_content(app.config['HEADER_TABS'],
--- a/app/static/widgets/color_picker.html
+++ b/app/static/widgets/color_picker.html
@@ -1,0 +1,575 @@
+<!-- Color Picker Widget -->
+<script>
+/*
+    All of the necesary color format conversion functions.
+    Taken from https://github.com/Qix-/color-convert/blob/master/conversions.js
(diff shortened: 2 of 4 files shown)


benbusby:
I'm a bit on the fence about this. I do like the idea of having a color picker embedded in Whoogle, but this seems like a fair amount of additional code that I'd need to maintain moving forward, and I'm trying to keep Whoogle as minimal as possible. Plus I think a lot of people who use color pickers (myself included) are fine with either using a browser extension or something like Firefox's built in eyedropper. It'd be nice if there was a way to load extensions from someone else's repo so that it could stay separate from Whoogle, but idk how well/if that would work.

Don-Swanson:
Thanks a ton for the contribution and the thoughtful implementation. After review we are going to pass on this one for now. A few reasons why:

1. Content Security Policy
Whoogle ships with a strict Content Security Policy to protect users. That means no inline scripts and very limited script sources. The widget would either require loosening the policy or ongoing work to thread nonces or hashes through templates and assets. We are not comfortable weakening Content Security Policy and we do not have bandwidth to maintain a complex policy for optional UI features.

2. Ongoing maintenance
Widgets that inject client UI tend to break whenever Google tweaks markup or when our parsing and templating change. Keeping this working across themes and mobile would add test surface and regression risk. That is real maintenance cost for something that is not core to private search.

3. Keep Whoogle lean
Project scope is privacy first, minimal server logic, and as little fluff as possible. Adding a built in color picker pushes us toward a heavier app that is outside that philosophy.

This would be a perfect fit for a browser extension or a small userscript. A content script can detect relevant queries and add a color picker overlay on the client side. That keeps all logic in the browser, preserves our Content Security Policy, and lets interested users opt in without changing the server. If you publish an extension or userscript, we are happy to point users to it in the docs.

We appreciate the effort and the idea. If we ever add a formal plugin hook that can coexist with our Content Security Policy and lean approach, we can revisit.

PR #1022, opened 2023-06-18
updated dependencies + Dependabot configured for pulling dependencies updates , but not auto accept

vecna666:


vecna666:
Why?

benbusby:
Sorry about that! I occasionally get PRs that have a bunch of merge commits that the user typically doesn't mean to open, and at a quick first glance this looked a lot like one of those. I generally just close those automatically, but should've looked closer at this one first. I can still pull in these changes though and tag you in the commit.

PR #1029, opened 2023-07-09
Added a search filter to restrict innapropriate searches

BlockmasterPlayz:
Added a search filter to restrict innapropriate words that may be searched so children will be safe when using whoogle... Can be changed in the future as I have not included all the possible words that could be entered...

### DIFF
--- a/app/filter.py
+++ b/app/filter.py
@@ -79,7 +79,15 @@ def clean_query(query: str) -> str:
     Returns:
         str: The query string without any "-site:..." filters
     """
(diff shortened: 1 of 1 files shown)


benbusby (changes requested):
This should probably be implemented to only be active when the safe search config option is enabled, rather than just removing all of these words by default. 

Also the restricted words list should be a standalone file that gets parsed when the app initializes, similar to translations, country lists, etc (see `app/static/settings`).

Moist-Cat:
Safe search should be more than enough to filter explicit content.

Issue #1122, opened 2024-02-20
[BUG] Translation uncaught exception


seirdotexe:
**Describe the bug**
 When a translatian isn't found, an uncaught exception is thrown.

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/flask/app.py", line 2190, in wsgi_app
    response = self.full_dispatch_request()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/flask/app.py", line 1486, in full_dispatch_request
    rv = self.handle_user_exception(e)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/flask/app.py", line 1341, in handle_user_exception
    return self.handle_http_exception(e)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/flask/app.py", line 1281, in handle_http_exception
    return self.ensure_sync(handler)(e)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/whoogle/app/routes.py", line 609, in page_not_found
    return render_template('error.html', error_message=str(e)), 404
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/flask/templating.py", line 151, in render_template
    return _render(app, template, context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/flask/templating.py", line 132, in _render
    rv = template.render(context)
         ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/jinja2/environment.py", line 1301, in render
    self.environment.handle_exception()
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/jinja2/environment.py", line 936, in handle_exception
    raise rewrite_traceback_stack(source=source)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/whoogle/app/templates/error.html", line 23, in top-level template code
[...]

Issue #1100, opened 2023-11-19
[BUG] Docker Container Going "Unhealthy" and Whoogle Stops working Every Day or Two


ilovepancakes95:
After upgrading to v0.8.4, about 1-2 days after starting the docker container, the status goes "Unhealthy" and Whoogle's webpage stops loading/container seemingly is completely not responding until I restart it. Then status is "Healthy" again until 1-2 days later where it fails yet again. The logs in docker for the container don't seem to show any errors. The behavior wasn't present when running the prior version before v0.8.4.

**Deployment Method**
- [ ] Heroku (one-click deploy)
- [x] Docker
- [ ] `run` executable
- [ ] pip/pipx
- [ ] Other: [describe setup]

**Version of Whoogle Search**
- [ ] Latest build from [source] (i.e. GitHub, Docker Hub, pip, etc)
- [X] Version 0.8.4
- [ ] Not sure

ryang3d:
I'm seeing a very similar (likely the same) issue on version 0.8.4 (manual install on Ubuntu 22.04, not docker). Service appears to be running fine but Whoogle becomes unreachable after roughly 24 hours and will not load up again until I manually restart the service. Like OP, this keeps happening every day or so, so requires a manual service restart every day or so too.

MaHivka:
I have an identical situation, except that the status "Unhealthy" appears after several search queries, I restart the container manually and everything works again (This is clearly not how it was planned to work). 
Version: Docker "latest" (most likely 0.8.4, the latest after all)
(p.s. I have installed new version, not updated it.)

thimplicity:
Same here unfortunately.

raindog2:
Same problem here:
Docker container latest version.

gravelfreeman:
Same thing here, I think Google is patching 3rd parties.

gravelfreeman:
Same thing. Here's all the logs related to a failed search.

```python
2023-12-20 12:53:19.806235-05:00ERROR:app:Exception on /autocomplete [POST]
2023-12-20 12:53:19.806349-05:00Traceback (most recent call last):
2023-12-20 12:53:19.806379-05:00File "/usr/local/lib/python3.11/site-packages/urllib3/connection.py", line 174, in _new_conn
2023-12-20 12:53:19.806398-05:00conn = connection.create_connection(
2023-12-20 12:53:19.806415-05:00^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1126, opened 2024-03-02
Fix uncaught exception "translation" from Jinja

ahmad-alkadri:
This (draft) PR potentially can fix Issues #1122 and (maybe) #1100.

Location of the bug found at the [app/routes.py](https://github.com/benbusby/whoogle-search/commit/7f0af097aa5724466f33f9e8c9b9ab9050a9b7b1#diff-f67826701212aab477be0634a23fdcd7ffdfe748b8ce35eb27b8f690d334c732) and [app/templates/error.html](https://github.com/benbusby/whoogle-search/commit/7f0af097aa5724466f33f9e8c9b9ab9050a9b7b1#diff-60a7913be4c694885dbfdfcd24057bc2b3f59f3edf0b0e01c31ecd46f40754be). 

Basically instead of querying the key from the `translation` dictionary at the template-level we should do it at the python func-level. 

Please click on the links to see the details of the commits.

Also modified one of the testing script.

Open to feedback.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ whoogle.env
 # vim
 *~
 *.swp
+
+# vscode
+.vscode/launch.json
--- a/app/routes.py
+++ b/app/routes.py
@@ -622,7 +622,7 @@ def internal_error(e):
     return render_template(
             'error.html',
             error_message='Internal server error (500)',
-            translation=translation,
+            continue_search_message=translation.get('continue-search', ''),
             farside='https://farside.link',
             config=g.user_config,
             query=urlparse.unquote(query),
--- a/app/templates/error.html
+++ b/app/templates/error.html
@@ -20,7 +20,7 @@
     </p>
     <hr>
     <p>
-        <h4><a class="link" href="https://farside.link">{{ translation['continue-search'] }}</a></h4>
+        <h4><a class="link" href="https://farside.link">{{ continue_search_message }}</a></h4>
         <ul>
             <li>
                 <a href="https://github.com/benbusby/whoogle-search">Whoogle</a>
--- a/test/test_results.py
+++ b/test/test_results.py
@@ -37,9 +37,11 @@ def test_get_results(client):
     assert rv._status_code == 200
 
     # Depending on the search, there can be more
-    # than 10 result divs
+    # than 5 result divs and less than 15 results
+    # within one single page. Should adjust this number
+    # accordingly.
     results = get_search_results(rv.data)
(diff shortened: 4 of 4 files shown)


ahmad-alkadri:
Also just realized that the test code failed at [this part](https://github.com/benbusby/whoogle-search/actions/runs/8121865884/job/22200698484?pr=1126#step:5:59) where:

```python
    def test_ddg_bang(client):
        # Bang at beginning of query
        rv = client.get(f'/{Endpoint.search}?q=!gh%20whoogle')
>       assert rv._status_code == 302
E       assert 200 == 302
E        +  where 200 = <WrapperTestResponse streamed [200 OK]>._status_code
```

So basically the endpoint responded with **200** (status ok successful) instead of **302**. Maybe we should modify the test script to include more scenarios. What do you think @benbusby?

benbusby:
@ahmad-alkadri the test error you saw should be fixed now -- it was due to DuckDuckGo changing their bang URL, so it needed to be updated to not use the version number anymore.

As far as the translation errors, it would probably be simpler to do something like the following:

```diff
- <input type="submit" id="search-submit" value="{{ translation['search'] }}">
+ <input type="submit" id="search-submit" value="{{ translation.get('search', 'Search') }}">
```

Calling the function directly in the html template means we could still just pass in the translation dict as before, but switch to using the `.get(key, default)` function in the template. With that approach, there would be little to no changes in the python code.

Also worth noting that the default value should always be a non-empty string, maybe just the default English translation of whichever element?

benbusby:
@ahmad-alkadri Actually never mind, I looked into this a bit further and it looks like the primary issue is that the error html template renders farside redirects when they aren't actually needed. I'm going to push a separate fix to make the redirect section only render if a query is present, otherwise it should just display the error and allow the user to return back to the home page.

benbusby:
Just pushed the change to `app/templates/error.html`, which resolves the original issue. Feel free to reopen this PR if you see anything that you think should still be refactored with the new approach!

ahmad-alkadri:
Hi @benbusby 

> the primary issue is that the error html template renders farside redirects

that is correct

> when they aren't actually needed. I'm going to push a separate fix to make the redirect section only render if a query is present, otherwise it should just display the error and allow the user to return back to the home page.

no problem with me. I've just pulled the main branch with your fixes and I saw that [my issue](https://github.com/benbusby/whoogle-search/issues/1122#issuecomment-1974697605)'s gone also :smile: Thanks!

PR #1129, opened 2024-03-21
changes proposed by pixee bot

bhaskarnn9:


benbusby (commented):
A lot of these changes aren't needed -- things like using a secure randomness generator for the UserAgent string isn't something that needs to be secure, or the secure subprocess call in `replit.py`, which is only ever called by Replit when Whoogle is deployed through them, and iirc doesn't have access to modules outside of what is available in base Python.

Also there seems to be a mix of using `requests.get` and `security.safe_requests.get`, and the `exception` module imported from `security.safe_requests` doesn't seem to exist (at least in v1.2.1).

With that said, I'm gonna close this PR since I don't think a lot of these changes are worth adding another dependency. I'll keep an eye on the `security` library from pixee though, seems like it could be more useful as it continues to get developed

PR #1131, opened 2024-03-22
Attempt to extract query on error condition and print trace

pantherman594:
When an error is hit, the search alternatives are not useful because the query is still encrypted. This PR tries to decrypt the search query for the error page. It also prints a stack trace of the exception for easier debugging.

### DIFF
--- a/app/routes.py
+++ b/app/routes.py
@@ -8,6 +8,8 @@ import re
 import urllib.parse as urlparse
 import uuid
 import validators
+import sys
+import traceback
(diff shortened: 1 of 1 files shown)


pantherman594:
I originally included the query with the stack trace to potentially help with debugging, but I realized logging these queries could be a privacy concern. Let me know if you think that would be useful to add back

benbusby (approved):
Looks good, agreed about not logging queries in the traceback

PR #1132, opened 2024-03-22
Custom bangs

### LINKED ISSUE #289: [FEATURE] Let the iBangs !i, !v, and !n go to Whoogle's own image, video and news search
Right now the iBangs !i, !v, and !n goes to duckduckgo. This is not bad but i think it should be better to integrate them with whoogle instead. And besides, image, video and news search isn't available with duckduckgo when javascript is off.

pantherman594:
Add the possibility for user-defined bangs, stored in `app/static/bangs`. These are parsed in alphabetical order, with the DDG bangs parsed first.

Closes  #289

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 venv/
+.venv/
 .idea/
 __pycache__/
 *.pyc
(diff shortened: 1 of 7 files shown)


benbusby (approved):
Works great, thanks!

PR #1141, opened 2024-04-21
Updating ad filter with Czech language version

AndrewNovikov:
If your version of Google is in Czech language, commercial links will not be filtered out because the word 'Sponzorováno' is missing in Ad keywords list BLACKLIST.


benbusby (approved):
Thanks!

PR #1183, opened 2024-10-07
Add ability to set custom dotenv path

tE3m:
Recently, I installed the project via `pipx` in combination with a dotenv file for the config as a  systemd service. However, the default (and hardcoded) location of the dotenv is the parent directory of the app, which makes sense if the app sits in the repo's directory. When installing with pipx though, one would need to put the dotenv file into the `site-packages` directory which doesn't seem sensible, even if harmless. As such, I thought it would make sense to allow setting a custom path to the dotenv (something like `~/.config/whoogle/whoogle.env`), which this change allows by passing .env's path to  the environment variable `WHOOGLE_DOTENV_PATH` while defaulting to the current behavior


benbusby (approved):
Looks good, thanks!

PR #1200, opened 2024-11-18
Update URL in README.md

klausagnoletti:
Updated URL pointing to how to set as default search engine so it actually works now.


benbusby (approved):
Thanks!

PR #1205, opened 2024-12-14
Reduce Docker image size by improving apk commands

PeterDaveHello:
Replace redundant apk flags `-U` and `--update` with `--no-cache` to prevent package index caching. This change unifies all apk package commands for better maintainability and smaller image size.


benbusby (approved):
Thank you!

Issue #1206, opened 2024-12-18
[BUG] Getting Sponsored results (french language)


Althior:
**Describe the bug**
Same issue as https://github.com/benbusby/whoogle-search/issues/1172, I have sponsored links showing in results. I guess there is no filtering for french sponsored links ("Sponsorisé" in french).

**To Reproduce**
1. Search anything
2. An ad could appear, always at the top

**Deployment Method**
- Docker

**Version of Whoogle Search**
- Latest build from Docker

**Desktop (please complete the following information):**
 - OS: Win10
 - Browser Firefox
 - Version 133

See example:
![389295031-8c92dbbe-0f32-4ecc-b795-b0057bf220f4](https://github.com/user-attachments/assets/a7cb71ab-b5a6-42ad-8fc3-5b291c7a9921)

Tell me if you need any information. Thank you!

TheRolfFR:
Hello fellow french compatriot, just do a PR for the same as https://github.com/benbusby/whoogle-search/commit/6924f5ce0d3987b79615079804489f03d6b2572f and add `Sponsorisé`.

FR: c'est assez simple à modifier mdr, fais juste la même chose

PR #1208, opened 2025-01-02
Update ad filter keyword for French language

Althior:
Fixes https://github.com/benbusby/whoogle-search/issues/1206


benbusby (approved):
Thank you!

PR #1212, opened 2025-01-17
Add UserAgent Options

Don-Swanson:
I've added a new config section that allows a user to choose between Lynx (The current workaround), Random (The Original method), or the ability to enter a custom User Agent String (in the event that Google decides to play whack-a-mole and kill the Lynx workaround, it's easier to choose a different user agent instead of rebuilding)

### DIFF
--- a/app/request.py
+++ b/app/request.py
@@ -78,7 +78,6 @@ def gen_user_agent(config, is_mobile) -> str:
 
     # If using custom user agent, return the custom string
     if config.user_agent == 'custom' and config.custom_user_agent:
-        print(f"Using custom user agent: {config.custom_user_agent}")  # Debug log
         return config.custom_user_agent
     
     # If using Lynx user agent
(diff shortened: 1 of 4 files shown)


app/request.py
```diff
@@ -198,24 +196,23 @@ def __init__(self, normal_ua, root_path, config: Config):
...
         send_tor_signal(Signal.HEARTBEAT)
 
-        self.language = (
-            config.lang_search if config.lang_search else ''
-        )
-
+        self.language = config.lang_search if config.lang_search else ''
         self.country = config.country if config.country else ''
-
+
```
xatier, line 201: nit: whitespace

benbusby (approved):
Thank you for looking into this, very much appreciated. I made a small change to include the link to the User Agent wiki page I just created, and removed some whitespace the other reviewer mentioned.

Issue #1230, opened 2025-06-04
[BUG] setting the alt reddit link to old.reddit.com causes the url to change to old.old.old.old.old. etc


RoyalOughtness:
**Describe the bug**

![Image](https://github.com/user-attachments/assets/4cdbf48e-abe8-45d7-a29d-abafba2242c5)

**To Reproduce**
1. Set WHOOGLE_ALT_RD=old.reddit.com
2. Search for something that results in a reddit link
3. click the link
4. observe that the url is https://old.old.old.old.old.old.old.old.old.old.reddit.com instead of https://old.reddit.com

**Deployment Method**
- [ ] Heroku (one-click deploy)
- [x] Docker
- [ ] `run` executable
- [ ] pip/pipx
- [ ] Other: [describe setup]

**Version of Whoogle Search**
- [ ] Latest build from [source] (i.e. GitHub, Docker Hub, pip, etc)
- [x] Version 0.9.3
- [ ] Not sure

allandiegoasilva:
@RoyalOughtness  Thanks for sharing your docker-compose.
I recommend simplifying the setup for initial testing — try running it without Traefik first to reduce variables and confirm the core functionality.
Once it's working properly, you can reintroduce Traefik and test the integration step-by-step.

Thank you,
Allan

Don-Swanson:
This has been incorporated into the Beta release: https://github.com/benbusby/whoogle-search/releases/tag/v1.0.0-beta

Please provide feedback if this is now resolved in beta.

PR #1235, opened 2025-07-16
fix: several bugs affecting alt links

RoyalOughtness:
This fixes a number of bugs:

1. This bug, https://github.com/benbusby/whoogle-search/issues/1230, where the subdomain is repeatedly reapplied to the same link. This was fixed by exiting the loop if the alt link is already present in the link. However, this is actually the symptom of a broader redundancy in the code: this loop [here](https://github.com/benbusby/whoogle-search/blob/main/app/filter.py#L652) calls get_site_alt for each link on the page, but then get_site_alt loops over all the site alts again [here](https://github.com/benbusby/whoogle-search/blob/main/app/utils/results.py#L198). In effect, this means the following is taking place:

```
foreach site alt
  foreach link
    foreach site alt
```

A broader refactor is needed to remove this redundancy/inefficiency. However, this PR contains the "exit if alt link is already present in the link" statement which is a working bandaid.

2. Adds edge case handling for simple.wikipedia.org which breaks if the provided alt link is a specific language of wikipedia.org

3. Fixes the link_desc field so that prefixes aren't ignored. Otherwise, we end up with link_descs like "www.https://old.reddit.com"

### DIFF
--- a/app/filter.py
+++ b/app/filter.py
@@ -1,5 +1,4 @@
 import cssutils
-from app.utils.misc import SKIP_PREFIX
 from bs4 import BeautifulSoup
 from bs4.element import ResultSet, Tag
 from cryptography.fernet import Fernet
--- a/app/utils/misc.py
+++ b/app/utils/misc.py
@@ -11,8 +11,6 @@ from bs4 import BeautifulSoup as bsoup
 from cryptography.fernet import Fernet
 from flask import Request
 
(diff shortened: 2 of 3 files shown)


Don-Swanson:
Closing in favor of https://github.com/benbusby/whoogle-search/releases/tag/v1.0.0-beta

PR #1237, opened 2025-08-23
Added a parsable JSON endpoint for results

### LINKED ISSUE #607: [FEATURE] Add a parsable (JSON or XML) endpoint for results

Current the `/search` endpoint returns a HTML file to display the results. This works well for UI users, which might be the first use case of Whoogle. But it's hard for programmatic use cases, as it requires parsing HTML.

I propose to add a new `/search.{json|xml}` endpoint to return the same results in a structured format: JSON or XML

### Idea 1: JSON

A `/search.json?q=%s` endpoint which would search results as JSON.

I don't think there's a standard JSON format to use, but we could take inspiration for https://zenserp.com/ or https://serpapi.com/.

### Idea 2: OpenSearch XML format

This project already uses opensearch, so we could leverage the same standard's response XML: https://github.com/dewitt/opensearch/blob/master/opensearch-1-1-draft-6.md#opensearch-response-elements

**Additional context**

I have a personal preference for JSON, as it seems to be today's web standard for APIs.

Adding parsable results would allow integration with various projects, via API (e.g. Zapier, Integromat), RSS or simple scripts. Some use cases include:
- keyword monitoring
- market research
- build a voice assistant
- and the imagination of the community

VeryUsual:
Fixes #607

### DIFF
--- a/app/routes.py
+++ b/app/routes.py
@@ -29,7 +29,6 @@ from app.utils.results import bold_search_terms,\
     add_currency_card, check_currency, get_tabs_content
 from app.utils.search import Search, needs_https, has_captcha
 from app.utils.session import valid_user_session
-from app.models.g_classes import GClasses
 from bs4 import BeautifulSoup as bsoup
 from flask import jsonify, make_response, request, redirect, render_template, \
     send_file, session, url_for, g


Don-Swanson:
@ProgramminCat 
The PR adds from app.models.g_classes import GClasses but never uses it in the code, is your implementation complete?

ProgramminCat:
Yes, my implementation is complete. I probably just used it before than stopped using it but never removed the import.

Don-Swanson:
In review and testing, I've noticed that the JSON endpoint doesn't follow the same Content Filtering and URL Sanitization (Anonymization) as the regular search results. To ensure that you receive similar results and outputs as what is provided via the UI, these should be implemented here as well. 

This is great progress though. A potential alternative that might be easier to implement and less work would be instead of parsing raw HTML, consider modifying the regular search endpoint to detect JSON requests (Accept: application/json). Then use the same filtered results but returning them in JSON format. This would ensure identical security measures for both results.

Don-Swanson:
Closing in favor of https://github.com/benbusby/whoogle-search/releases/tag/v1.0.0-beta

Issue #1239, opened 2025-09-10
Searching no longer work


ArrowNemesis:
Using Chrome Version 139.0.7258.154 says: 

"Update your browser
Your browser isn't supported any more. To continue your search, upgrade to a recent version."

hobby-grade:
I am also experiencing this issue and think it may be related to the user agent change to allow searches to work earlier this year when Google stopped Javascript-free searches. 

<img width="892" height="376" alt="Image" src="https://github.com/user-attachments/assets/821736e3-99b9-4345-bdd4-13be03a46be3" />

jrg72:
Jumping on this also, I'm locked out in the same way from using Whoogle. I'm using Zen (essentially Firefox) but I also doubt that's relevant, as I think this is a Google change.

Don-Swanson:
As part of that fix i enabled the ability to change the UA from the Configuration dropdown on the Search page. I can confirm that there are UAs that do still work even though the Lynx no longer works. I afraid to post what does work though because I’m sure the big G is monitoring this and will kill off UAs we publicly list as working.

Don-Swanson:
@hobby-grade 
The config change I’m referring to is on the Whoogle Search Home page that you can live on demand change without having to change the UA on the backend. 

If you are using that option on the Search page drop down, and it’s not working on mobile, it must be something with your browser or UA. The one I configured on my desktop, I just tried on my mobile and it works. Maybe you used a UA that Google knows would never be Mobile? Not sure

LupinThidr:
I mentioned this in #1211 earlier this year (buried in a sea of "+1" and "same here"):
> If Lynx ever gets blocked (probably will, if the current game of whack-a-mole that's been going on against YouTube is anything to go by), a few more UA's worth trying are the Nintendo ones (3DS / Wii) and Opera Mini's. There's still millions of users in poorer countries reliant on Opera Mini and its serverside rendering.

Other embedded / lower spec devices they can't just suddenly kill might also be worth investigating such as UAs corresponding to built in browsers in older phones, smart TVs.

blackletum:
> I've could get whoogle working on one of my instance by setting a UA via Configuration in the UI. But i can't get it working on my other instance by setting it as a env var in my `docker-compose.yaml`
> 
> services:
>   whoogle-search:
>     image: benbusby/whoogle-search:latest
>     **snip**
>     environment:
>       **snip**
[...]

PR #1240, opened 2025-09-10
Updated UI Config to respect User Agent in ENV file if exists

Don-Swanson:
To fix issue #1239 

Updated UI Config to respect User Agent in ENV file if exists

### DIFF
--- a/app/models/config.py
+++ b/app/models/config.py
@@ -37,8 +37,12 @@ def get_rule_for_selector(stylesheet: CSSStyleSheet,
 
 class Config:
     def __init__(self, **kwargs):
-        # User agent configuration
-        self.user_agent = kwargs.get('user_agent', 'LYNX_UA')
+        # User agent configuration - default to env_conf if environment variables exist, otherwise default
+        env_user_agent = os.getenv('WHOOGLE_USER_AGENT', '')
+        env_mobile_agent = os.getenv('WHOOGLE_USER_AGENT_MOBILE', '')
+        default_ua_option = 'env_conf' if (env_user_agent or env_mobile_agent) else 'default'
+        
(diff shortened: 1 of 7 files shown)


benbusby (approved):

kraetos10:
will you be able to create a new release so docker can pickup the changes ?

Hydranet:
@Don-Swanson Does this also respect the variable when running the container using  a systemd-unit. I'm running Whoogle with podman and have a systemd-unit to start my container which looks like this.

```
[Unit]
Description=Whoogle

[Container]
AutoUpdate=registry
ContainerName=whoogle
Image=docker.io/benbusby/whoogle-search:latest
Network=host
#PublishPort=5000:5000
Environment=WHOOGLE_FALLBACK_ENGINE_URL=https://duckduckgo.com/?k1=-1&q=
Environment=WHOOGLE_CONFIG_THEME=dark
Environment=WHOOGLE_USER_AGENT="Chrome/140.0.0.0 Safari/537.36"
```
However still getting the same as the issue that's the problem mentioned in issue #1239. Could also be that this user agent doesn't work but just want to verify first whether this respects only what's mentioned the env section of a composer.yml file or that it respects the environment variable in general?

Don-Swanson:
> > It shouldn't matter what underlying system you are running it on. However, I can almost guarantee that any semi modern Chrome/Safari UA string will NOT work.
> 
> Thanks for verifying that, I'll continue to look for UA that does work.

Try just updating your container image and use the default (Aka leave off that ENV Variable) 😉

Hydranet:
> Try just updating your container image and use the default (Aka leave off that ENV Variable) 😉

It wasn't working this morning, but now it works again without me having to change UA. Thanks! :)

PR #1245, opened 2025-09-17
Use httpx with HTTP/2 for faster response

rstefko:
In my case with HTTP/1.1 each search request took always approximately 3 seconds. With HTTP/2 its just milliseconds.

### DIFF
--- a/app/request.py
+++ b/app/request.py
@@ -11,6 +11,7 @@ from stem import Signal, SocketError
 from stem.connection import AuthenticationFailure
 from stem.control import Controller
 from stem.connection import authenticate_cookie, authenticate_password
+import httpx
 
 MAPS_URL = 'https://maps.google.com/maps'
 AUTOCOMPLETE_URL = ('https://suggestqueries.google.com/'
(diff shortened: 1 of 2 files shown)


Don-Swanson:
I love this idea, but there appears to be a few issues.

Redirect handling changes
requests.get follows redirects by default. httpx.Client does not. Your new call does not pass follow_redirects=True, so flows that relied on 30x chains to land on the final Google page may now receive a redirect response body instead. That can break parsing and captcha detection logic. The httpx docs are explicit about this default. Set follow_redirects=True to keep parity. 

Tor and SOCKS support breaks
Whoogle supports Tor via a SOCKS proxy. httpx supports SOCKS only if the extra is installed. The PR adds httpx[http2] but not httpx[socks]. If self.proxies contains socks5h://..., httpx without SOCKS support will fail or bypass the proxy depending on setup. Add httpx[socks] to requirements or gate HTTPX use when SOCKS is configured. 

Dependency hardening for recent CVEs
HTTPX depends on httpcore which uses h11 for HTTP/1.1. There is a 2025 h11 vulnerability fixed by updating stacks to h11 0.16.0 and httpcore 1.0.9. Pin or minimum-bound those to avoid vulnerable resolver combos in older envs. 

API stability of the proxies argument
You pinned httpx[http2]==0.27.2, where proxies= still works, but it was removed in 0.28. If anyone upgrades independently, Client(..., proxies=...) will error. Prefer proxy= or proper mounts.

rstefko:
All issues should be addressed. Proxies do not work good with HTTP/2, so I use HTTP/2 it only when we have a direct connection to Google.

Don-Swanson:
Thanks. Your PR has sparked an idea/effort to replace Python Requests entirely with httpx throughout the code. 
Allow me a little time to test this theory first and see if it's better to replace it throughout.

PR #1250, opened 2025-09-28
Looks like we can re-enable view_image functionality

rstefko:


Don-Swanson:
Hi @rstefko 
Thank you for this. Would you mind re-submitting this against the beta branch?
I would like to add this to the pending beta release, but would like for you to receive credit.

PR #1276, opened 2025-11-24
Revert leta

Don-Swanson:
In this release, we reverted the Mullvad Leta implementation. It served us well and we’re grateful for it, but it’s time to let it take its final voyage to Valhalla. In its place, we added a feature that can randomly generate UA strings based on patterns we found in the ones that still work. You can also bring your own rotating list or stick with a static UA if that’s your thing. There’s a config toggle that shows the active UA at the bottom of the results to help with debugging. On top of that, we added two small tools that let you generate offline UA lists and automatically test them. The second of which was instrumental in finding the working string patterns we’re using now.

### DIFF
--- a/LETA_INTEGRATION.md
+++ b/LETA_INTEGRATION.md
@@ -1,0 +1,137 @@
+# Mullvad Leta Backend Integration
+
+## Overview
+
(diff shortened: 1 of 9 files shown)


blackletum:
When is it planned to bring this PR into main?

Don-Swanson:
> When is it planned to bring this PR into main?

On Wednesday after some opportunity for people to test the beta. Check the Releases for the beta.

